PR #2, opened 2019-10-22
Replay events to integrations after successful initialization

prabrisha-rudder:
Wait for a specified time for the enabled integrations to get loaded. Then only replay the events to the integrations that are successfully initialized and not to the integrations that are failed to get initialized as for the failed integrations, the global queue of corresponding integration keeps getting incremented for adding the events and not able to send to the destinations.


sayan-rudder:
👍

PR #5, opened 2019-11-15
Namespace refactor

prabrisha-rudder:


sayan-rudder:
will update cdn link after final review

PR #20, opened 2020-01-18
Replace console.log with logger.debug

masad-frost:
The console gets spammed with `anonymousId: ...`

I just replaced `console.log` with `logger.debug` so that it's only logged during development


sayan-rudder:
Thanks

PR #58, opened 2020-05-04
Small improvements to the GA integration

pierreavizou:
This PR tries to bring a few improvements to the GA native integration: 

- Allow setting the `allowLinker` property on the tracker from config. This defaults to false (default GA setting), until configuration is exposed to user in the config plane. `allowLinker` is used by GA for cross-domain tracking.

- Fix sending userId to GA when calling `identify()`. Effectively sends user-generated userId or anonymousId if none is defined. Brings consistency with cloud-mode integration.

- Sets `userId` on the tracker even if `identify()` has not been called during this session. Allow retrieving a previously-saved userId on fullpage reload, without needing to call `identify()` on each page load. Brings consistency with cloud mode, where userId is sent with every request.

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/58)


sayan-rudder:
👍

PR #102, opened 2020-07-20
Send the events to server/destination that are made only after load call

prabrishac:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/102)


sayan-mitra:
👍

PR #114, opened 2020-07-30
added bugsnag destination

ruchiramoitra:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/114)

### DIFF
--- a/integrations/Bugsnag/browser.js
+++ b/integrations/Bugsnag/browser.js
@@ -1,7 +1,7 @@
 import logger from "../../utils/logUtil";
-import  ScriptLoader  from "../ScriptLoader";
+import ScriptLoader from "../ScriptLoader";
 
- class Bugsnag {
+class Bugsnag {
   constructor(config) {
     this.releaseStage = config.releaseStage;
     this.apiKey = config.apiKey;
@@ -9,20 +9,20 @@ import  ScriptLoader  from "../ScriptLoader";
     this.setIntervalHandler = undefined;
   }
 
-   init() {
+  init() {
     logger.debug("===in init Bugsnag===");
     ScriptLoader(
       "bugsnag-id",
       "https://d2wy8f7a9ursnm.cloudfront.net/v6/bugsnag.min.js"
     );
 
-     this.setIntervalHandler = setInterval(
+    this.setIntervalHandler = setInterval(
       this.initBugsnagClient.bind(this),
       1000
     );
   }
 
-   initBugsnagClient() {
+  initBugsnagClient() {
     if (window.bugsnag !== undefined) {
       window.bugsnagClient = window.bugsnag(this.apiKey);
       window.bugsnagClient.releaseStage = this.releaseStage;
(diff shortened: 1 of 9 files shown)


integrations/Bugsnag/browser.js
```diff
@@ -0,0 +1,59 @@
...
+    );
+
+     this.setIntervalHandler = setInterval(
+      this.initBugsnagClient.bind(this),
+      1000
+    );
+  }
+
+   initBugsnagClient() {
+    if (window.bugsnag !== undefined && window.bugsnag !== void 0) {
```
sayan-mitra, line 30: what does void 0 does?
can we use a corresponding easy readable syntax

integrations/index.js
```diff
@@ -35,6 +36,7 @@ const integrations = {
   FACEBOOK_PIXEL: FBPixel.default,
   LOTAME: Lotame.default,
   OPTIMIZELY: Optimizely.default,
+  BUGSNAG: Bugsnag.default
```
sayan-mitra, line 39: also add a corresponding entry in cname file. 
can you also check after making the change that selective destination for sending events is working fine for bugsnag

integrations/Bugsnag/browser.js
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+   identify(rudderElement) {
+    let traits = rudderElement.message.context.traits;
+    let traitsFinal = {
+      id: rudderElement.message.userId,
+      name: traits.name,
+      email: traits.email
+    };
+    console.log(traitsFinal);
```
sayan-mitra, line 54: change to logger.debug

integrations/Bugsnag/browser.js
```diff
@@ -0,0 +1,59 @@
...
+
+   isReady() {
+    logger.debug("in bugsnag isReady");
+    return !!window.bugsnagClient;
+  }
+
+   identify(rudderElement) {
+    let traits = rudderElement.message.context.traits;
+    let traitsFinal = {
+      id: rudderElement.message.userId,
```
sayan-mitra, line 50: should we do rudderElement.message.userId || rudderElement.message.anonymousId ?

integrations/Bugsnag/browser.js
```diff
@@ -0,0 +1,59 @@
...
+    logger.debug("in bugsnag isLoaded");
+    return !!window.bugsnagClient;
+  }
+
+   isReady() {
+    logger.debug("in bugsnag isReady");
+    return !!window.bugsnagClient;
+  }
+
+   identify(rudderElement) {
```
sayan-mitra, line 47: also check for page/track etc api calls, the sdk is not crashing.
Ideally it should be logging the error in console and not crash, still check once

integrations/Bugsnag/browser.js
```diff
@@ -0,0 +1,59 @@
+import logger from "../../utils/logUtil";
+import  ScriptLoader  from "../ScriptLoader";
+import {
```
sayan-mitra, line 3: please remove unused imports

integrations/Bugsnag/browser.js
ruchiramoitra: Similarly it was done when initialising other destinations. To check if undefined or not.. Should I remove it ?

integrations/Bugsnag/browser.js
ruchiramoitra: yes. added it.

integrations/Bugsnag/browser.js
ruchiramoitra: no called page and track the sdk is not crashing

integrations/Bugsnag/browser.js
ruchiramoitra: removed

integrations/Bugsnag/browser.js
ruchiramoitra: changed

sayan-mitra:
👍

PR #120, opened 2020-08-18
added tvsquared as destination

ruchiramoitra:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/120)

### DIFF
--- a/integrations/ScriptLoader.js
+++ b/integrations/ScriptLoader.js
@@ -6,9 +6,13 @@ const ScriptLoader = (id, src) => {
   js.src = src;
   js.async = true;
   js.type = "text/javascript";
+  if(id === "TVSquared-integration"){
+    js.defer = true;
+  }
   js.id = id;
   const e = document.getElementsByTagName("script")[0];
   logger.debug("==script==", e);
+  console.log("==script==", e)
   e.parentNode.insertBefore(js, e);
 };
 
--- a/integrations/ScriptLoader.js
+++ b/integrations/ScriptLoader.js
@@ -6,13 +6,9 @@ const ScriptLoader = (id, src) => {
   js.src = src;
   js.async = true;
   js.type = "text/javascript";
-  if(id === "TVSquared-integration"){
-    js.defer = true;
-  }
   js.id = id;
   const e = document.getElementsByTagName("script")[0];
   logger.debug("==script==", e);
-  console.log("==script==", e)
   e.parentNode.insertBefore(js, e);
 };
 
--- a/integrations/TVSquared/browser.js
+++ b/integrations/TVSquared/browser.js
@@ -18,14 +18,8 @@ class TVSquared {
     let url = document.location.protocol === "https:" ? "https://" : "http://";
     url += `collector-${this.clientId}.tvsquared.com/`;
     window._tvq.push(["setSiteId", this.brandId]);
-    window._tvq.push(["setTrackerUrl", `&${url}tv2track.php`]);
+    window._tvq.push(["setTrackerUrl", `${url}tv2track.php`]);
     ScriptLoader("TVSquared-integration", `${url}tv2track.js`);
-
-    window._tvq.push([
-      () => {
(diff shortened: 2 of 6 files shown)


integrations/TVSquared/browser.js
```diff
@@ -0,0 +1,104 @@
...
+    window._tvq.push([
+      function () {
+        this.deleteCustomVariable(5, "page");
+      },
+    ]);
+  }
+
+  isLoaded() {
+    logger.debug("in TVSqaured isLoaded");
+    return !!window_tvq;
```
prabrishac, line 29: typo? should be !!window._tvq

should we not do something like - 
return !!(window._tvq && window._tvq.push !== Array.prototype.push);

integrations/TVSquared/browser.js
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  isReady() {
+    logger.debug("in TVSqaured isReady");
+
+    return !!window_tvq;
```
prabrishac, line 35: same as  `isLoaded`

integrations/TVSquared/browser.js
```diff
@@ -0,0 +1,104 @@
...
+          action[key] = value;
+        }
+      }
+    }
+    window._tvq.push([
+      function () {
+        this.setCustomVariable(5, "session", JSON.stringify(session), "visit");
+      },
+    ]);
+    if (event !== "Response") {
```
prabrishac, line 89: or `All response`
Should we make the check case insensitive?

integrations/integration_cname.js
```diff
@@ -42,6 +42,7 @@ const commonNames = {
   FULLSTORY: "FULLSTORY",
   Fullstory: "FULLSTORY",
   BUGSNAG: "BUGSNAG",
+  TVSQUARED: "TVSQUARED",
```
prabrishac, line 45: also put corresponding entry in `client_server_name.js`

integrations/ScriptLoader.js
```diff
@@ -6,9 +6,13 @@ const ScriptLoader = (id, src) => {
   js.src = src;
   js.async = true;
   js.type = "text/javascript";
+  if(id === "TVSquared-integration"){
+    js.defer = true;
+  }
```
prabrishac, line 11: currently we load all js as `async` only. remove the `defer` attribute now. we have the task of making all the scripts with `defer` in pipeline. Will do the change as part of that task.

integrations/ScriptLoader.js
```diff
@@ -6,9 +6,13 @@ const ScriptLoader = (id, src) => {
...
   js.id = id;
   const e = document.getElementsByTagName("script")[0];
   logger.debug("==script==", e);
+  console.log("==script==", e)
```
prabrishac, line 15: remove all `console.log`

integrations/ScriptLoader.js
ruchiramoitra: ok

integrations/TVSquared/browser.js
ruchiramoitra: yes it was a type changed it.

integrations/TVSquared/browser.js
ruchiramoitra: All response was not mentioned in the sdk doc so did not include it. Doing the case insensitive part

integrations/integration_cname.js
ruchiramoitra: integrations/client_server_name.js

I have put here..

integrations/TVSquared/browser.js
ruchiramoitra: done

integrations/integration_cname.js
prabrishac: ohh... missed it.

arnab-p (approved):

PR #144, opened 2020-09-17
Lytics destination

ruchiramoitra:
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/144)


integrations/Lytics/browser.js
```diff
@@ -0,0 +1,141 @@
...
+  constructor(config) {
+    this.accountId = config.accountId;
+    this.stream = config.stream;
+    this.blockload = config.blockload;
+    this.loadid = config.loadid;
+    this.name = "LYTICS";
+  }
+
+  loadLyticsScript() {
+    !(function () {
```
sayan-mitra, line 31: We can keep either **!** or **(func call)**
https://stackoverflow.com/questions/3755606/what-does-the-exclamation-mark-do-before-the-function

integrations/Lytics/browser.js
```diff
@@ -0,0 +1,141 @@
...
+      src: `//c.lytics.io/api/tag/${this.accountId}/latest.min.js`,
+    });
+  }
+
+  init() {
+    this.loadLyticsScript();
+    logger.debug("===in init Lytics===");
+  }
+
+  isLoaded = () => {
```
sayan-mitra, line 100: it's better to use one syntax for the entire file. we can skip doing arrow functions.

integrations/Lytics/browser.js
```diff
@@ -0,0 +1,141 @@
...
+    return !!(window.jstag && window.jstag.push !== Array.prototype.push);
+  };
+
+  identify(rudderElement) {
+    logger.debug("in Lytics identify");
+    // eslint-disable-next-line camelcase
+    const user_id =
+      rudderElement.message.userId || rudderElement.message.anonymousId;
+    const { traits } = rudderElement.message.context;
+    const payload = { user_id, ...traits };
```
sayan-mitra, line 119: - does Lytics support nested traits?
- what if traits is empty like {}

integrations/Lytics/browser.js
```diff
@@ -0,0 +1,141 @@
...
+    window.jstag.send(this.stream, payload);
+  }
+
+  page(rudderElement) {
+    logger.debug("in Lytics page");
+    // eslint-disable-next-line no-underscore-dangle
+    const _e = rudderElement.message.name;
```
sayan-mitra, line 126: is there a significance of naming variables this way?

integrations/Lytics/browser.js
```diff
@@ -0,0 +1,141 @@
...
+    const payload = { _e, ...properties };
+    window.jstag.send(this.stream, payload);
+  }
+
+  track(rudderElement) {
+    logger.debug("in Lytics track");
+    // eslint-disable-next-line no-underscore-dangle
+    const _e = rudderElement.message.event;
+    const { properties } = rudderElement.message;
+    const payload = { _e, ...properties };
```
sayan-mitra, line 137: same as traits comment

sayan-mitra (commented):
Let's also add destination specific test files in the same destination folder.
like: 
 ```
|- integrations
         |- Lytics
                |- source code and test files
```

File: integrations/Lytics/browser.js (hunk 1 above)
Comment (ruchiramoitra) [reply integrations/Lytics/browser.js:31]: Done

File: integrations/Lytics/browser.js (hunk 2 above)
Comment (ruchiramoitra) [reply integrations/Lytics/browser.js:100]: I replaced the previous function with error function and keeping this as it is as per the aslant rules.

File: integrations/Lytics/browser.js (hunk 4 above)
Comment (ruchiramoitra) [reply integrations/Lytics/browser.js:126]: a _e named property will be created in Lytics under which the name will be shown. As segment is saving them under this property name we are doing the same. 
_e is used in Lytics to save the event name.

File: integrations/Lytics/browser.js (hunk 2 above)
Comment (ruchiramoitra) [reply integrations/Lytics/browser.js:100]: changed

File: integrations/Lytics/browser.js (hunk 3 above)
Comment (ruchiramoitra) [reply integrations/Lytics/browser.js:119]: 1. 

yes it supports nested properties. The nested properties are sent by flattening the json. if array of json object is sent for example: 
company :[{"id":"abc123","name":{"second_name":"xyz","first_name":"abc"}},{"id":"abc1234","name":{"second_name":"xyza","first_name":"abcd"}}]

Lytics will send it as: 
company_len: 2
company_json: [{"id":"abc123","name":{"second_name":"xyz","first_name":"abc"}},{"id":"abc1234","name":{"second_name":"xyza","first_name":"abcd"}}]
company.id: abc123
company.name.second_name: xyz
company.name.first_name: abc
company.id: abc1234
company.name.second_name: xyza
company.name.first_name: abcd


2. If traits is empty only user_id will be sent

File: integrations/Lytics/browser.js (hunk 5 above)
Comment (ruchiramoitra) [reply integrations/Lytics/browser.js:137]: same as above

File: integrations/Lytics/browser.js (hunk 4 above)
Comment (sayan-mitra) [reply integrations/Lytics/browser.js:126]: feels a bit out of place.. lets have explicit key in payload
like : {"_e": $event_name}

File: integrations/Lytics/browser.js (hunk 3 above)
Comment (sayan-mitra) [reply integrations/Lytics/browser.js:120]: 👍

File: integrations/Lytics/browser.js (hunk 4 above)
Comment (ruchiramoitra) [reply integrations/Lytics/browser.js:127]: Changed

File: integrations/Lytics/browser.js [hunk 6]
```diff
@@ -0,0 +1,138 @@
...
+    })();
+
+    // Define config and initialize Lytics tracking tag.
+
+    window.jstag.init({
+      loadid: this.loadid,
+      blocked: this.blockload,
+      stream: this.stream,
+      sessecs: 1800,
+      src: `//c.lytics.io/api/tag/${this.accountId}/latest.min.js`,
```
humanchimp, line 92: We recommend using an absolute URL:
```suggestion
      src: `https://c.lytics.io/api/tag/${this.accountId}/latest.min.js`,
```

integrations/Lytics/browser.js
```diff
@@ -0,0 +1,138 @@
...
+      rudderElement.message.userId || rudderElement.message.anonymousId;
+    const { traits } = rudderElement.message.context;
+    const payload = { user_id, ...traits };
+    window.jstag.send(this.stream, payload);
+  }
+
+  page(rudderElement) {
+    logger.debug("in Lytics page");
+    const { properties } = rudderElement.message;
+    const payload = { _e: rudderElement.message.name, ...properties };
```
humanchimp, line 127: The Lytics JS tag has a `.pageView()` method for sending a page view event into Lytics. It works by sending a payload with the special key `"_e": "pv"`. Using `"_e"` for custom events is not recommended because it can get overwritten. For instance, if the tag is configured using this configuration:

```javascript
jstag.config = {
  send: {
    pageViewOnFirst: true
  }
};
```

Then the `"_e"` event will get overwritten by the string `"pv"` which may cause problems with data mapping.

We recommend sending custom events with the key `"event"` instead, which will also be mapped as a user event in the standard LQL.

If you are just trying to track a page view event in Lytics, you may want to consider use the built-in `jstag.pageView(payload)` method here instead of `jstag.send(payload)`

humanchimp (commented):
Hello, I am an engineer at Lytics. I left some notes about your PR below. Thanks!

integrations/Lytics/browser.js
ruchiramoitra: Changed it.

integrations/Lytics/browser.js
ruchiramoitra: Thanks. Changed it.

arnab-p (changes requested):
Please rebase

sayan-mitra:
closing this as #165 addresses the same.

PR #164, opened 2020-11-24
Pendo sdk

Shivam4415:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [x] Lint rules pass locally
- [x] The code changed/added as part of this pull request has been covered with tests
- [x] All tests related to the changed code pass in development

### Code review 

- [x]  This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [x] Changes have been reviewed by at least one other engineer

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/164)

### DIFF
--- a/integrations/Pendo/browser.js
+++ b/integrations/Pendo/browser.js
@@ -1,7 +1,5 @@
-/* eslint-disable no-param-reassign */
-/* eslint-disable no-underscore-dangle */
 /* eslint-disable class-methods-use-this */
-/* eslint-disable lines-between-class-members */
+/* eslint-disable no-underscore-dangle */
 import logger from "../../utils/logUtil";
 import ScriptLoader from "../ScriptLoader";
 
@@ -25,6 +23,7 @@ class Pendo {
     );
     logger.debug("===in init Pendo===");
 
+    // Make a call to initPendo until pendoCli gets references to pendo object.
     this.setIntervalHandler = setInterval(this.initPendo.bind(this), 1000);
   }
 
@@ -32,6 +31,8 @@ class Pendo {
     if (this.isReady()) {
       window.pendoCli = window.pendo;
       clearInterval(this.setIntervalHandler);
+
+      // After pendo is initialized set debugger if debugMode is enabled.
       if (this.isDebugMode) {
         this.enableDebugging();
       } else {
@@ -39,6 +40,7 @@ class Pendo {
       }
     }
   }
+
   isLoaded() {
     logger.debug("in PENDO isLoaded");
     return !!window.pendoCli;
@@ -73,30 +75,33 @@ class Pendo {
   }
 
   recurse(cur, prop, result) {
+    const res = result;
     if (Object(cur) !== cur) {
-      result[prop] = cur;
+      res[prop] = cur;
     } else if (Array.isArray(cur)) {
       const l = cur.length;
       for (let i = 0; i < l; i += 1)
-        this.recurse(cur[i], prop ? `${prop}.${i}` : `${i}`);
-      if (l === 0) result[prop] = [];
+        this.recurse(cur[i], prop ? `${prop}.${i}` : `${i}`, res);
+      if (l === 0) res[prop] = [];
     } else {
       let isEmpty = true;
       Object.keys(cur).forEach((key) => {
         isEmpty = false;
-        this.recurse(cur[key], prop ? `${prop}.${key}` : key, result);
+        this.recurse(cur[key], prop ? `${prop}.${key}` : key, res);
       });
-      if (isEmpty) result[prop] = {};
+      if (isEmpty) res[prop] = {};
     }
-    return result;
+    return res;
   }
 
   flattenAgentPayload(data) {
     return this.recurse(data, "", {});
   }
 
+  /* Once iser is identified Pendo makes Track call to track user activity.
+   */
   track(rudderElement) {
-    const props = rudderElement.message;
+    const props = rudderElement.message.properties;
     window.pendoCli.track(rudderElement.message.event, props);
   }
 
@@ -119,6 +124,10 @@ class Pendo {
     this._identify(obj);
   }
 
+  /* Pendo's identify call works intelligently, once u have identifioed a visitor/user,
+   *or associated a visitor to a group/account then Pendo save this data in local storage and
+   *any further upcoming calls are done taking user infor from local.
+   */
   _identify(obj) {
     window.pendoCli.identify(obj);
   }
--- a/integrations/Pendo/browser.js
+++ b/integrations/Pendo/browser.js
@@ -81,17 +81,13 @@ class Pendo {
     let visitorObj = {};
     let accountObj = {};
     const { groupId } = this.analytics;
-    const { traits } = rudderElement.message.context;
     const id =
-      rudderElement.message.userId ||
-      this.constructPendoAnonymousId(rudderElement.message.anonymousId);
-    visitorObj.id = id;
-    if (traits) {
-      visitorObj = {
-        ...visitorObj,
-        ...traits,
-      };
-    }
+      this.analytics.userId ||
+      this.constructPendoAnonymousId(this.analytics.anonymousId);
+    visitorObj = {
+      id,
+      ...this.analytics.userTraits,
+    };
 
     if (groupId) {
       accountObj = { id: groupId, ...this.analytics.groupTraits };
@@ -107,14 +103,11 @@ class Pendo {
     let accountObj = {};
     let visitorObj = {};
     const { userId, traits } = rudderElement.message;
-    accountObj.id =
-      rudderElement.message.groupId || rudderElement.message.anonymousId;
-    if (traits) {
-      accountObj = {
-        ...accountObj,
-        ...traits,
-      };
-    }
+    accountObj.id = this.analytics.groupId || this.analytics.anonymousId;
+    accountObj = {
+      ...accountObj,
+      ...traits,
+    };
 
     if (userId) {
       visitorObj = {
--- a/integrations/Pendo/browser.js
+++ b/integrations/Pendo/browser.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-param-reassign */
 /* eslint-disable no-underscore-dangle */
 /* eslint-disable class-methods-use-this */
 /* eslint-disable lines-between-class-members */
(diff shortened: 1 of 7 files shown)


integrations/Pendo/browser.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  isReady() {
+    return window.pendo && window.pendo.isReady && window.pendo.isReady();
+  }
+
+  identify(rudderElement) {
+    let visitorObj = {};
+    visitorObj.id =
+      rudderElement.message.userId || rudderElement.message.anonymousId;
```
ruchiramoitra, line 53: this is correct.. What segment does is I user id is not present they append the anonymous id with _PENDO_T_ . We can implement this too

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,94 @@
...
+    if (rudderElement.message.context.traits) {
+      visitorObj = {
+        ...visitorObj,
+        ...rudderElement.message.context.traits,
```
ruchiramoitra, line 57: please check if nested objects (the traits) are sent if it is being handled by panda or we have to send it as flattened?

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,94 @@
...
+      };
+    }
+
+    this._identify({ visitor: visitorObj });
+  }
+
+  track(rudderElement) {
+    window.pendoCli.track(rudderElement.message.event);
```
ruchiramoitra, line 65: I checked track events are sent like this in panda sdk --- >
 pendo.track("User Registered", {
    userId: user.id,
    plan: user.plan,
    accountType: "Facebook"
  });
});

So shouldn't the properties go along with the name of the event? 
Here also check about the nested properties.

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,94 @@
...
+    if (rudderElement.message.context.traits) {
+      accountObj = {
+        ...accountObj,
+        ...rudderElement.message.context.traits,
+      };
+    }
+    this._identify({ account: accountObj });
+  }
+
+  _identify(obj) {
```
ruchiramoitra, line 81: As sayan da had commented in the pond doc : 

can you check if this helps: https://github.com/segmentio/analytics.js-integrations/blob/master/integrations/pendo/lib/index.js#L124

Are we doing this? if not what is the difference between these two calls?

integrations/index.js
```diff
@@ -40,7 +41,7 @@ const integrations = {
   OPTIMIZELY: Optimizely.default,
   BUGSNAG: Bugsnag.default,
   FULLSTORY: Fullstory.default,
-  TVSQUARED: TVSquared.default,
```
ruchiramoitra, line 43: This should not be deleted?

Shivam4415: fixed

integrations/Pendo/browser.js
Shivam4415: Ok will do that, and what u are suggesting is actually correct, it will be easy to understand anonymous user by looking at Prefix :: PENDO_T

integrations/Pendo/browser.js
Shivam4415: Yes I am doing same what Sayan da suggested.

integrations/Pendo/browser.js
Shivam4415: fixed

integrations/Pendo/browser.js
Shivam4415: Sending properties along with track calls.
We don't need to flatten this properties. Pendo take care of nested objects.
Fixed

integrations/Pendo/browser.js
Shivam4415: Flattened the traits object and sending it to PendoAgent.
Pendo has an agent section that accepts traits that are sent for that visitor or account.

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,144 @@
...
+    if (rudderElement.message.context.traits) {
+      visitorObj = {
+        ...visitorObj,
+        ...this.flattenAgentPayload(rudderElement.message.context.traits),
+      };
+    }
+
+    this._identify({ visitor: visitorObj });
+  }
+
```
ruchiramoitra, line 76: lets keep this after the identify group calls and give a comment to explain. Or we could take this functionality to utils file and import...

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,144 @@
...
+    return !message.userId;
+  }
+
+  identify(rudderElement) {
+    let visitorObj = {};
+    const id = this.isUserAnonymous(rudderElement.message)
+      ? this.constructPendoAnonymousId(rudderElement.message.anonymousId)
+      : rudderElement.message.userId;
+    visitorObj.id = id;
+    if (rudderElement.message.context.traits) {
```
ruchiramoitra, line 67: let { traits } = rudderElement.message.context... 
before if so that we do not need to get the value twice and just use the variable.

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  flattenAgentPayload(data) {
+    return this.recurse(data, "", {});
+  }
+
+  /* Once iser is identified Pendo makes Track call to track user activity.
+   */
+  track(rudderElement) {
+    const props = rudderElement.message.properties;
```
ruchiramoitra, line 104: check if props is null what will happen Nothing will break right?

ruchiramoitra: also keep a check for event... if its not present throw a message maybe that event field is not present so track call not called.

integrations/Pendo/browser.js
Shivam4415: Fixed

integrations/Pendo/browser.js
Shivam4415: after the identify group calls./....
I did not get what it means.

integrations/Pendo/browser.js
Shivam4415: Nothing breaks if prop is null.
Fixed null check for event name.

integrations/Pendo/browser.js
Shivam4415: fixed

ruchiramoitra (approved):
LGTM

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,140 @@
...
+    this.name = "PENDO";
+    this.setIntervalHandler = undefined;
+    logger.debug("Config ", config);
+  }
+
+  init() {
+    window.pendo = {};
+    window.pendoCli = undefined;
+    window.pendo.apiKey = this.apiKey;
+    ScriptLoader(
```
sayan-mitra, line 21: why not use the initialisation script from here: https://developers.pendo.io/docs/?bash#introduction ?
also their script seems similar to us where the 'initialize','identify','updateOptions','pageLoad' are already present even when the script is not ready. (Also  include track there in the list as I think it got supported in later version). Then our isReady()  and isLoaded() can always return true

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,140 @@
...
+    );
+    logger.debug("===in init Pendo===");
+
+    // Make a call to initPendo until pendoCli gets references to pendo object.
+    this.setIntervalHandler = setInterval(this.initPendo.bind(this), 1000);
+  }
+
+  initPendo() {
+    if (this.isReady()) {
+      window.pendoCli = window.pendo;
```
sayan-mitra, line 33: check earlier comment to decide if this is necessary

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,140 @@
...
+      : rudderElement.message.userId;
+    visitorObj.id = id;
+    if (traits) {
+      visitorObj = {
+        ...visitorObj,
+        ...flattenJsonPayload(traits),
+      };
+    }
+
+    this._identify({ visitor: visitorObj });
```
sayan-mitra, line 97: is this indirection required? we can put the line here directly

utils/utils.js
```diff
@@ -420,6 +420,30 @@ function getUserProvidedConfigUrl(configUrl) {
   return url;
 }
 
+function recurse(cur, prop, result) {
```
sayan-mitra, line 423: Add a comment explaining what this method does

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,140 @@
...
+
+  /* Pendo's identify call works intelligently, once u have identified a visitor/user,
+   *or associated a visitor to a group/account then Pendo save this data in local storage and
+   *any further upcoming calls are done taking user info from local.
+   * To track user perndo maps user to Visitor in Pendo.
+   */
+  identify(rudderElement) {
+    let visitorObj = {};
+    const { traits } = rudderElement.message.context;
+    const id = this.isUserAnonymous(rudderElement.message)
```
sayan-mitra, line 86: use a shorthand || ?

integrations/Pendo/browser.js
sayan-mitra, line 97: add account obj too? for ex: a group call is made earlier than a later/second identify.

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /*
+   *Group call maps to an account for which visitor belongs.
+   *It is same as identify call but here we send account object.
+   */
+  group(rudderElement) {
+    const obj = {};
+    let accountObj = {};
+    const { traits } = rudderElement.message.context;
```
sayan-mitra, line 118: our group level traits are present at top level.
user traits are present inside context.
I think we need to send both visitor and account object by reading traits from top level and under context.

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,140 @@
...
+    accountObj.id =
+      rudderElement.message.groupId || rudderElement.message.anonymousId;
+    if (traits) {
+      accountObj = {
+        ...accountObj,
+        ...flattenJsonPayload(traits),
```
sayan-mitra, line 124: also is flatten necessary?

integrations/Pendo/browser.js
```diff
@@ -0,0 +1,140 @@
...
+      };
+      obj.account = accountObj;
```
sayan-mitra, line 126: for account there seems to be a concept of parentAccounts. Can you check that

sayan-mitra (changes requested):
Few comments on functionality.
Add a PR to production-staging branch too

utils/utils.js
Shivam4415: Fixed

integrations/Pendo/browser.js
Shivam4415: For a single line its is not necessary, removed indirection to private function.

Shivam4415: If adding account object means I am trying to explicity associate a group to a user/visitor, but Pendo does it implicitly for me if there is already an account associated.
Its not a problem to do explicitly but what I think let Pendo handle it.

integrations/Pendo/browser.js
Shivam4415: Added shorthand || and removed function isUserAnonymous.

Shivam4415:
> Few comments on functionality.
> Add a PR to production-staging branch too

integrations/Pendo/browser.js
Shivam4415: Initialized scripts in init function.

integrations/Pendo/browser.js
Shivam4415: This line of code is not required anymore.

integrations/Pendo/browser.js
Shivam4415: Added account object if present

integrations/Pendo/browser.js
Shivam4415: Sending both object

integrations/Pendo/browser.js
Shivam4415: Removing flatten as Pendo cannot recognize property with dot operator

sayan-mitra (approved):

sayan-mitra:
closing this as #167 addresses this

Issue #282, opened 2021-07-28
Facebook Pixel contentType ignored


alan-cooney:
In https://github.com/rudderlabs/rudderstack-docs/blob/master/destinations/advertising/fb-pixel.md it is stated that you can set contentType like this:

```
rudderanalytics.track("Product Added", {
        //...
    },{
        'Facebook Pixel': { contentType: 'mycustomtype' }
    }
);
```

However this doesn't work, as there is a bug in https://github.com/rudderlabs/rudder-sdk-js/blob/production/integrations/FacebookPixel/browser.js whereby a presumably legacy "options" parameter doesn't exist.

It looks like we can change:

```
const { options, properties } = rudderElement.message;
```

To:

```
const { properties } = rudderElement.message;
const options = rudderElement?.integrations?.[clientToServerNames.FACEBOOK_PIXEL];
```

alan-cooney:
https://github.com/rudderlabs/rudder-sdk-js/pull/283

saikumarrs:
PR #283 has been merged. Hence, closing the issue.

PR #283, opened 2021-07-28
Fix Facebook content type

alan-cooney:
## Description of the change

Fix to allow contentType to be set on Facebook Events (e.g. if you want to send a "destination" or "flight" rather than "product" content type.

## Type of change

## Related issues

> Fix [#282]()

### DIFF
--- a/integrations/FacebookPixel/browser.js
+++ b/integrations/FacebookPixel/browser.js
@@ -499,7 +499,8 @@ class FacebookPixel {
   }
 
   getContentType(rudderElement, defaultValue) {
-    const { options, properties } = rudderElement.message;
+    const { properties } = rudderElement.message;
+    const options = rudderElement?.message?.integrations?.FACEBOOK_PIXEL;
     if (options && options.contentType) {
       return [options.contentType];
     }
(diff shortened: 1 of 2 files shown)


integrations/FacebookPixel/browser.js
```diff
@@ -499,7 +499,8 @@ class FacebookPixel {
   }
 
   getContentType(rudderElement, defaultValue) {
-    const { options, properties } = rudderElement.message;
+    const { properties } = rudderElement.message;
+    const options = rudderElement?.message?.integrations?.FACEBOOK_PIXEL;
```
ruchiramoitra, line 503: we don't need the key `options`, instead just the `contentType` from rudderElement?.message?.integrations?.FACEBOOK_PIXEL

alan-cooney: Yes absolutely - we can simplify the whole function quite a bit here actually.

alan-cooney: I've updated this all and tried to do some manual testing for the various use cases. Let me know what you think.

ruchiramoitra (approved):

PR #400, opened 2021-12-22
JS SDK v1 with beacon feature

MoumitaM:
## Description of the change

> Added browsers sendBeacon feature in our v1 JS SDK. Customers now can use a single SDK to send data either via xhr or browser's sendBeacon. Default is xhr. If the load() call has the "useBeacon" parameter set to true then only the data is sent via beacon.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/400)

### DIFF
--- a/analytics.js
+++ b/analytics.js
@@ -60,7 +60,7 @@ function enqueue(rudderElement, type) {
   if (!this.eventRepository) {
     this.eventRepository = EventRepository;
   }
-  this.eventRepository.enqueue(rudderElement, type);
+  this.eventRepository.enqueue(rudderElement, type, this.beaconPlugin);
 }
 
 /**
@@ -98,7 +98,7 @@ class Analytics {
     };
     this.loaded = false;
     this.loadIntegration = true;
-    this.useBeacon = false;
+    this.beaconPlugin = undefined;
   }
 
   /**
@@ -1014,7 +1014,27 @@ class Analytics {
     }
 
     if (options && options.useBeacon === true) {
-      ScriptLoader("beaconPlugin", BEACON_PLUGIN_URL);
+      const pluginName = "beaconQueue";
+      ScriptLoader(pluginName, BEACON_PLUGIN_URL);
+      const interval = setInterval(
+        function () {
+          if (window.hasOwnProperty(pluginName)) {
+            const intMod = window[pluginName];
+            clearInterval(interval);
+            try {
+              this.beaconPlugin = new intMod();
+              this.eventRepository.useBeacon = options.useBeacon;
+            } catch (e) {
+              logger.error(pluginName, " initialization failed", e);
+            }
+          }
+        }.bind(this),
+        100
+      );
+
+      setTimeout(() => {
+        clearInterval(interval);
+      }, MAX_WAIT_FOR_INTEGRATION_LOAD);
     }
 
     function errorHandler(error) {
--- a/analytics.js
+++ b/analytics.js
@@ -42,7 +42,6 @@ import logger from "./utils/logUtil";
 import { addDomEventHandlers } from "./utils/autotrack.js";
 import ScriptLoader from "./integrations/ScriptLoader";
 import parseLinker from "./utils/linker";
-import { isNotEmpty } from "./integrations/utils/commonUtils";
 
 const queryDefaults = {
   trait: "ajs_trait_",
@@ -941,10 +940,7 @@ class Analytics {
     if (options && options.secureCookie) {
       storageOptions = { ...storageOptions, secure: options.secureCookie };
     }
-
-    if (isNotEmpty(storageOptions)) {
-      this.storage.options(storageOptions);
-    }
+    this.storage.options(storageOptions);
 
     if (options && options.integrations) {
       Object.assign(this.loadOnlyIntegrations, options.integrations);
--- a/analytics.js
+++ b/analytics.js
@@ -986,11 +986,7 @@ class Analytics {
       this.loadIntegration = !!options.loadIntegration;
     }
 
-    this.eventRepository.writeKey = writeKey;
-    if (serverUrl) {
-      this.eventRepository.url = serverUrl;
-    }
-    this.eventRepository.initialize(options);
+    this.eventRepository.initialize(writeKey, serverUrl, options);
     this.initializeUser();
     this.setInitialPageProperties();
     this.loaded = true;
--- a/analytics.js
+++ b/analytics.js
@@ -43,6 +43,7 @@ import logger from "./utils/logUtil";
 import { addDomEventHandlers } from "./utils/autotrack.js";
 import ScriptLoader from "./integrations/ScriptLoader";
 import parseLinker from "./utils/linker";
+import { isNotEmpty } from "./integrations/utils/commonUtils";
 
 const queryDefaults = {
   trait: "ajs_trait_",
@@ -930,12 +931,23 @@ class Analytics {
       throw Error("failed to initialize");
     }
 
+    let storageOptions = {};
     if (options && options.logLevel) {
       logger.setLogLevel(options.logLevel);
     }
+
     if (options && options.setCookieDomain) {
-      this.storage.options({ domain: options.setCookieDomain });
+      storageOptions = { ...storageOptions, domain: options.setCookieDomain };
     }
+
+    if (options && options.secureCookie) {
+      storageOptions = { ...storageOptions, secure: options.secureCookie };
+    }
+
+    if (isNotEmpty(storageOptions)) {
+      this.storage.options(storageOptions);
+    }
+
     if (options && options.integrations) {
       Object.assign(this.loadOnlyIntegrations, options.integrations);
       tranformToRudderNames(this.loadOnlyIntegrations);
--- a/analytics.js
+++ b/analytics.js
@@ -27,12 +27,13 @@ import {
   transformToServerNames,
   checkReservedKeywords,
   getReferrer,
-  getReferringDomain
+  getReferringDomain,
 } from "./utils/utils";
 import {
   CONFIG_URL,
   MAX_WAIT_FOR_INTEGRATION_LOAD,
-  INTEGRATION_LOAD_CHECK_INTERVAL
+  INTEGRATION_LOAD_CHECK_INTERVAL,
+  BEACON_PLUGIN_URL,
 } from "./utils/constants";
 import { integrations } from "./integrations";
 import RudderElementBuilder from "./utils/RudderElementBuilder";
@@ -45,7 +46,7 @@ import parseLinker from "./utils/linker";
 
 const queryDefaults = {
   trait: "ajs_trait_",
-  prop: "ajs_prop_"
+  prop: "ajs_prop_",
 };
 
 // https://unpkg.com/test-rudder-sdk@1.0.5/dist/browser.js
@@ -93,10 +94,11 @@ class Analytics {
     this.readyCallback = () => {};
     this.executeReadyCallback = undefined;
     this.methodToCallbackMapping = {
-      syncPixel: "syncPixelCallback"
+      syncPixel: "syncPixelCallback",
     };
     this.loaded = false;
     this.loadIntegration = true;
+    this.useBeacon = false;
   }
 
   /**
@@ -169,7 +171,7 @@ class Analytics {
         if (destination.enabled) {
           this.clientIntegrations.push({
             name: destination.destinationDefinition.name,
-            config: destination.config
+            config: destination.config,
           });
         }
       }, this);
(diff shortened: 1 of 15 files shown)


analytics.js
```diff
@@ -60,7 +60,7 @@ function enqueue(rudderElement, type) {
   if (!this.eventRepository) {
     this.eventRepository = EventRepository;
   }
-  this.eventRepository.enqueue(rudderElement, type);
+  this.eventRepository.enqueue(rudderElement, type, this.beaconPlugin);
```
saikumarrs, line 63: ?

analytics.js
```diff
@@ -1023,6 +1023,11 @@ class Analytics {
       }
     }
 
+    if (options && options.useBeacon === true) {
+      this.eventRepository.useBeacon = options.useBeacon;
+      this.eventRepository.initializeTransportMechanism();
+    }
```
saikumarrs, line 1029: This should be done before queuing is started (Ln 985)

analytics.js
```diff
@@ -1023,6 +1023,11 @@ class Analytics {
       }
     }
 
+    if (options && options.useBeacon === true) {
+      this.eventRepository.useBeacon = options.useBeacon;
+      this.eventRepository.initializeTransportMechanism();
```
saikumarrs, line 1028: This should be internally handled by the event repository when it is triggered to start queuing. SDK should not be aware of these details.

utils/EventRepository.js
```diff
@@ -47,6 +48,8 @@ class EventRepository {
     this.url = "";
     this.state = "READY";
     this.batchSize = 0;
+    this.useBeacon = false;
+    this.beaconQueue = beaconQueue;
```
saikumarrs, line 52: For Ln 63, why is `payloadQueue` initialized by default?

saikumarrs (changes requested):

utils/EventRepository.js
```diff
@@ -227,12 +230,30 @@ class EventRepository {
...
     // modify the url for event specific endpoints
     const url = this.url.slice(-1) == "/" ? this.url.slice(0, -1) : this.url;
-    // add items to the queue
-    this.payloadQueue.addItem({
-      url: `${url}/v1/${type}`,
-      headers,
-      message,
-    });
+    if (this.useBeacon) {
+      const targetUrl = `${url}/beacon/v1/batch`;
```
saikumarrs, line 234: Can you verify if we have to send it to the Beacon batch endpoint or the regular batch endpoint works?

utils/storage/beaconQueue.js
```diff
@@ -0,0 +1,113 @@
...
+
+  flushQueue(headers, batch, url, writekey) {
+    headers.type = "application/json";
+    batch.map((event) => {
+      event.sentAt = new Date().toISOString();
+    });
+    const data = { batch };
+    const payload = JSON.stringify(data, this.replacer);
+    const blob = new Blob([payload], headers);
+    const isPushed = navigator.sendBeacon(`${url}?writeKey=${writekey}`, blob);
```
saikumarrs, line 87: I believe we are sending the write key to the batch endpoint for some reason. Please find this out from the server team and the possibility to replace it with the regular endpoint.

utils/storage/beaconQueue.js
```diff
@@ -0,0 +1,113 @@
...
+
+const defaults = {
+  queue: "queue",
+  maxPayloadSize: 64 * 1000,
+};
+
+class BeaconQueue {
+  constructor() {
+    this.storage = Store;
+    this.maxItems = 10;
```
saikumarrs, line 13: Does this have to be a hard code value?

saikumarrs (changes requested):

utils/EventRepository.js
```diff
@@ -227,12 +230,30 @@ class EventRepository {
...
+  }
+
+  initializeTransportMechanism() {
+    const sendQueueData = this.sendQueueDataForBeacon.bind(this);
+    const url = this.url.slice(-1) == "/" ? this.url.slice(0, -1) : this.url;
+    const targetUrl = `${url}/beacon/v1/batch`;
+    this.beaconQueue.url = targetUrl;
+    this.beaconQueue.writekey = this.writeKey;
+    window.addEventListener("unload", sendQueueData);
   }
```
saikumarrs, line 257: Why do these have to be here than in the beacon lib?

analytics.js
MoumitaM: By mistake, removed it

analytics.js
MoumitaM: This is used to clear the previous queue if any and destroy the queue. The name is not appropriate.

utils/storage/beaconQueue.js
MoumitaM: You mean the maxItem?

analytics.js
saikumarrs: Why are we starting the queue in the first place when the decision to use the beacon or not is not yet decided?

utils/EventRepository.js
MoumitaM: Yes can be moved to beacon lib.

utils/EventRepository.js
MoumitaM: This is for xhr, haven’t changed anything.

analytics.js
MoumitaM: I think there is some misunderstanding. Say, first time you load the page with useBeacon: true. That time a queue will be created with the timestamp. Then if you reload the page a new queue will be created. So to send data from the previous queue and destroy it this is necessary. 
I have updated the name to "initializeBeaconQueue".

analytics.js
saikumarrs: Appreciate the quick response but can you respond after you push the changes? I can re-verify and resolve it.

utils/EventRepository.js
saikumarrs: We're supposed to change the implementation to initialize either XHR or Beacon based on the load options.

utils/storage/beaconQueue.js
saikumarrs: Yes

utils/EventRepository.js
MoumitaM: done

utils/storage/beaconQueue.js
MoumitaM: done

utils/EventRepository.js
MoumitaM: done

analytics.js
MoumitaM: done

analytics.js
MoumitaM: done

utils/storage/beaconQueue.js
```diff
@@ -0,0 +1,130 @@
...
+    this.writekey = "";
+    this.queueName = `${defaults.queue}.${Date.now()}`;
+  }
+
+  sendQueueDataForBeacon() {
+    this.sendDataFromQueueAndDestroyQueue();
+  }
+
+  init(url, writekey, options) {
+    if (options.maxItems) this.maxItems = options.maxItems;
```
saikumarrs, line 28: `options.maxItems` sounds vague. Let's do `options.beaconQueue.maxItems`

utils/storage/beaconQueue.js
```diff
@@ -0,0 +1,130 @@
...
+
+  clearTimer() {
+    if (this.timeOutActive) {
+      clearTimeout(this.flushQueueTimeOut);
+      this.timeOutActive = false;
+    }
+  }
+}
+const beaconQueue = new BeaconQueue();
+export default beaconQueue;
```
saikumarrs, line 130: Why are we creating the instance by default? Can't we create more than one queue?

utils/xhrModule.js
```diff
@@ -0,0 +1,110 @@
...
+    this.payloadQueue.addItem({
+      url: `${url}/v1/${type}`,
+      headers,
+      message,
+    });
+  }
+}
+
+const xhr = new XHR();
+export default xhr;
```
saikumarrs, line 110: Same query as beacon queue. Why do we have to create all the instances by default?

utils/EventRepository.js
```diff
@@ -226,14 +69,16 @@ class EventRepository {
...
 
     // modify the url for event specific endpoints
-    const url = this.url.slice(-1) == "/" ? this.url.slice(0, -1) : this.url;
-    // add items to the queue
-    this.payloadQueue.addItem({
-      url: `${url}/v1/${type}`,
-      headers,
-      message,
-    });
+    const url = this.url.slice(-1) === "/" ? this.url.slice(0, -1) : this.url;
```
saikumarrs, line 72: Why does this have to be done for each queue item?

utils/EventRepository.js
```diff
@@ -226,14 +69,16 @@ class EventRepository {
...
+    if (this.useBeacon) {
+      const targetUrl = `${url}/beacon/v1/batch`;
```
saikumarrs, line 74: Set the target URL during queue init and get rid of the `useBeacon` variable

analytics.js
```diff
@@ -1001,6 +990,7 @@ class Analytics {
     if (serverUrl) {
```
saikumarrs, line 990: This check is redundant. Also, take writeKey and serverUrl along with the options in the `initialize` method.

utils/storage/beaconQueue.js
MoumitaM: done

utils/xhrModule.js
```diff
@@ -0,0 +1,110 @@
...
+
+const queueOptions = {
+  maxRetryDelay: 360000,
+  minRetryDelay: 1000,
+  backoffFactor: 2,
+  maxAttempts: 10,
+  maxItems: 100,
+};
+
+class XHR {
```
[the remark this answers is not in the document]
MoumitaM: done

utils/EventRepository.js
MoumitaM: done

utils/EventRepository.js
MoumitaM: done

analytics.js
MoumitaM: done

utils/xhrModule.js
MoumitaM: Please check the Ln. 39
`xhrQueue.processQueueElement`
I am not sure how to call `processQueueElement` class fn from inside another function 
tried with `this.processQueueElement`.

utils/storage/beaconQueue.js
MoumitaM: why do we need more than one queue, that will complicate things more?

MoumitaM: Done

utils/xhrModule.js
MoumitaM: done, found a workaround.

utils/EventRepository.js
```diff
@@ -41,162 +18,41 @@ class EventRepository {
...
-        );
+  initialize(writeKey, url, options) {
+    let queueOptions = {};
+    let targetUrl;
+    this.writeKey = writeKey;
+    this.url = url.slice(-1) === "/" ? url.slice(0, -1) : url;
+    if (options && options.useBeacon) {
+      if (
+        options &&
+        options.beaconQueue &&
```
saikumarrs, line 35:
```suggestion
        options.beaconQueueOptions &&
```

utils/EventRepository.js
```diff
@@ -225,15 +81,9 @@ class EventRepository {
...
 
-    // modify the url for event specific endpoints
-    const url = this.url.slice(-1) == "/" ? this.url.slice(0, -1) : this.url;
-    // add items to the queue
-    this.payloadQueue.addItem({
-      url: `${url}/v1/${type}`,
-      headers,
-      message,
-    });
+    this.queue.enqueue(headers, message, type);
```
saikumarrs, line 84: I don't think the headers defined in Ln67 are applicable for Beacon.

saikumarrs: You can move the headers preparation responsibility to the queues.

utils/EventRepository.js
```diff
@@ -41,162 +18,41 @@ class EventRepository {
    *Creates an instance of EventRepository.
    * @memberof EventRepository
    */
-  constructor(options) {
-    this.eventsBuffer = [];
+  constructor() {
     this.writeKey = "";
     this.url = "";
```
saikumarrs, line 23: These two needn't be defined at the class level

utils/EventRepository.js
MoumitaM: Used in BeaconQueue.js --> fn flushQueue --> const blob = new Blob([payload], headers);

saikumarrs: Beacon queue ignores the headers I believe, which is why the write key is being passed as a query parameter. Can you please verify that it is part of the original implementation?

MoumitaM: Yes it is part of the original implementation. But not needed. new Blob([payload], headers) - only takes the data and the type. Which in this case is application/json.

saikumarrs (approved):

PR #436, opened 2022-01-19
posthog loaded function 

shrouti1507:
## Description of the change

> Description here

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/436)

### DIFF
--- a/integrations/Posthog/browser.js
+++ b/integrations/Posthog/browser.js
@@ -60,6 +60,8 @@ class Posthog {
       }, e.__SV = 1)
     }(document, window.posthog || []);
 
+    const { POSTHOG } = this.analytics.loadOnlyIntegrations;
+
     const configObject = {
       api_host: this.yourInstance,
       autocapture: this.autocapture,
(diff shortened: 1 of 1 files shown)


integrations/Posthog/browser.js
```diff
@@ -68,6 +70,10 @@ class Posthog {
       property_blacklist: this.propertyBlackList,
       disable_cookie: this.disableCookie,
     };
+
+    if (POSTHOG && POSTHOG.loaded) {
+      configObject.loaded = POSTHOG.loaded;
+    }
```
ItsSudip, line 76: We can add a check if loaded is a function or not, as we are expecting it as a function. Otherwise, it looks good.

ItsSudip (approved):

integrations/Posthog/browser.js
saikumarrs: Sounds good. @shrouti1507 please take care of it.

saikumarrs: But let's do that only if PostHog fails to initialize if the `loaded` parameter is not a function. I would assume PostHog would already have this validation.

shrouti1507: Checked sending `page` events to `posthog`. When we set a function inside `loaded` for example `posthog.identify('1234')` , it is identifying the user along with creating a page view. Now, when we set `loaded: "ABC"`, it is creating a single page view ( without throwing any error). That means, posthog is internally discarding the erroneous format of loaded variable. @ItsSudip

ItsSudip: Ok, then we don't need to validate it as posthog is already discarding from their end.

PR #455, opened 2022-02-14
Updating packages and removing dependencies

annez:
* No more lodash
* Upgrading storejs to use latest
* Adding in filesize plugin to see sizes on build

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/455)


dios-david (approved):

saikumarrs:
Hi @annez!
Many of your suggestions are already incorporated into the current production version. We are working on replacing the `lodash` dependencies, which contribute significantly to the size.

Once again, thanks for all the suggestions!

PR #466, opened 2022-02-25
Identify traits update for Facebook Pixel.

ItsSudip:
## Description of the change

We used to send the traits as it was, without any mapping. Now we are mapping them in the same way as to cloud mode.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/466)


integrations/FacebookPixel/constants.js
```diff
@@ -6,4 +6,72 @@ const CNameMapping = {
...
+      "traits.id",
+      "context.traits.userId",
+      "context.traits.id",
+      "anonymousId",
+    ],
+  },
+  {
+    destKey: "em",
+    sourceKeys: ["traits.email", "context.traits.email"],
+  },
```
shrouti1507, line 24: Previously for web device mode we used to fetch data from `context.traits` field only. We have changed the order of priority here. Let us check and be confirmed.

MoumitaM (dismissed):

MoumitaM:
@ItsSudip, Change the base from production to production-staging

integrations/FacebookPixel/constants.js
shrouti1507: @ItsSudip we can omit `traits` JS sdk only can expect `context.traits`  in identify.

integrations/FacebookPixel/browser.js
```diff
@@ -76,10 +77,30 @@ class FacebookPixel {
...
       const traits = rudderElement.message.context
         ? rudderElement.message.context.traits
         : undefined;
-      window.fbq("init", this.pixelId, traits);
+      // here we are sending other traits apart from the reserved ones.
+      reserve.forEach((element) => {
+        delete traits[element];
+      });
+      payload = { ...payload, ...traits };
+      window.fbq("init", this.pixelId, payload);
```
saikumarrs, line 103: With these changes, the final payload looks quite different than before. Wouldn't this be a breaking change for users?
CC @utsabc

ItsSudip: I don't think there is any use of other attributes apart from [these](https://developers.facebook.com/docs/meta-pixel/advanced/advanced-matching).

saikumarrs: The sample payloads sent to FB SDK are as follows:
Before:
```
{
  "firstName": "John",
  "lastName": "Cena",
  "email": "john@wwe.com",
  "phone": "",
  "gender": "m",
  "birthday": "19990101",
  "city": "London",
  "country": "England",
  "zip": "81929120",
  "foo": "bar"
}
```

After:
```
{
  "external_id": "user001",
  "em": "john@wwe.com",
  "ph": "",
  "ge": "m",
  "db": "19990101",
  "ln": "Cena",
  "fn": "John",
  "ct": "London",
  "zp": "81929120",
  "country": "England",
  "foo": "bar"
}
```

Clearly, the field names are quite different and there is a good possibility that the users are affected if they were relying on the previously sent fields.
We need to make sure that the changes are backward compatible.
CC @utsabc

integrations/FacebookPixel/browser.js
```diff
@@ -76,10 +77,30 @@ class FacebookPixel {
...
+        "city",
+        "birthday",
+        "gender",
+        "street",
+        "zip",
+        "country",
+      ];
+      // this construcPayload will help to map the traits in the same way as cloud mode
+      let payload = constructPayload(rudderElement.message, traitsMapper);
       const traits = rudderElement.message.context
```
utsabc, line 95: Use `or` instead of ternary?

integrations/FacebookPixel/browser.js
utsabc: @ItsSudip We should create a flag for backward compatibility. `Use Updated Mapping` by default it should be kept false also we should say that we recommend migrating as the earlier one will be depreciating

integrations/FacebookPixel/util.js
```diff
@@ -0,0 +1,21 @@
+/* eslint-disable no-param-reassign */
+import get from "get-value";
+
+const constructPayload = (object, mapper) => {
```
utsabc, line 4: Lets keep this in common utils with proper documentation (inline)

integrations/FacebookPixel/browser.js
utsabc: @ItsSudip We should keep a config in the web app for backward compatibility. For example `use updated mapping` which should be false by default when enabled the above mapping will be triggered.
Also let's inform in a callout that we will depreciate the existing mapping soon

utsabc (approved):

utsabc (approved):

saikumarrs (approved):

PR #497, opened 2022-04-13
Feature: Auto capture anonymous id 

MoumitaM:
## Description of the change

> For the customers migrating from Segment, SDK will auto capture Segment’s anonymousID and set it as RS’s anonymousID.
 - Added an option in the load call, `autoAssignAnonId`. If you set it to `segment`, it will auto capture the anonymousID if present in cookie or local storage.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/497)

### DIFF
--- a/analytics.js
+++ b/analytics.js
@@ -106,7 +106,7 @@ class Analytics {
   /**
    * initialize the user after load config
    */
-  initializeUser(autoAssignAnonId) {
+  initializeUser(anonymousIdOptions) {
     this.userId =
       this.storage.getUserId() != undefined ? this.storage.getUserId() : "";
 
@@ -123,7 +123,7 @@ class Analytics {
         ? this.storage.getGroupTraits()
         : {};
 
-    this.anonymousId = this.getAnonymousId(autoAssignAnonId);
+    this.anonymousId = this.getAnonymousId(anonymousIdOptions);
 
     // save once for storing older values to encrypted
     this.storage.setUserId(this.userId);
@@ -936,9 +936,9 @@ class Analytics {
     this.storage.clear(flag);
   }
 
-  getAnonymousId(autoAssignAnonId) {
+  getAnonymousId(anonymousIdOptions) {
     // if (!this.loaded) return;
-    this.anonymousId = this.storage.getAnonymousId(autoAssignAnonId);
+    this.anonymousId = this.storage.getAnonymousId(anonymousIdOptions);
     if (!this.anonymousId) {
       this.setAnonymousId();
     }
@@ -1072,7 +1072,7 @@ class Analytics {
     }
 
     this.eventRepository.initialize(writeKey, serverUrl, options);
-    this.initializeUser(options ? options.autoAssignAnonId : undefined);
+    this.initializeUser(options ? options.anonymousIdOptions : undefined);
     this.setInitialPageProperties();
     this.loaded = true;
     if (
--- a/analytics.js
+++ b/analytics.js
@@ -1072,11 +1072,7 @@ class Analytics {
     }
 
     this.eventRepository.initialize(writeKey, serverUrl, options);
-    if (options && options.autoAssignAnonId) {
-      this.initializeUser(options.autoAssignAnonId);
-    } else {
-      this.initializeUser();
-    }
+    this.initializeUser(options ? options.autoAssignAnonId : undefined);
     this.setInitialPageProperties();
     this.loaded = true;
     if (
--- a/analytics.js
+++ b/analytics.js
@@ -106,7 +106,7 @@ class Analytics {
   /**
    * initialize the user after load config
    */
-  initializeUser() {
+  initializeUser(autoAssignAnonId) {
     this.userId =
       this.storage.getUserId() != undefined ? this.storage.getUserId() : "";
 
@@ -123,7 +123,7 @@ class Analytics {
         ? this.storage.getGroupTraits()
         : {};
 
-    this.anonymousId = this.getAnonymousId();
+    this.anonymousId = this.getAnonymousId(autoAssignAnonId);
 
     // save once for storing older values to encrypted
     this.storage.setUserId(this.userId);
@@ -936,9 +936,9 @@ class Analytics {
     this.storage.clear(flag);
   }
 
-  getAnonymousId() {
+  getAnonymousId(autoAssignAnonId) {
     // if (!this.loaded) return;
-    this.anonymousId = this.storage.getAnonymousId();
+    this.anonymousId = this.storage.getAnonymousId(autoAssignAnonId);
     if (!this.anonymousId) {
       this.setAnonymousId();
     }
@@ -1072,7 +1072,11 @@ class Analytics {
     }
 
     this.eventRepository.initialize(writeKey, serverUrl, options);
-    this.initializeUser();
+    if (options && options.autoAssignAnonId) {
+      this.initializeUser(options.autoAssignAnonId);
+    } else {
+      this.initializeUser();
+    }
     this.setInitialPageProperties();
     this.loaded = true;
     if (
--- a/dist/rudder-sdk-js/index.d.ts
+++ b/dist/rudder-sdk-js/index.d.ts
@@ -50,6 +50,16 @@ declare module "rudder-sdk-js" {
   }
 
   /**
+   * Represents the options parameter for anonymousId
+   */
+  interface anonymousIdOptions {
+    autoCapture?: {
+      enabled?: boolean;
+      source?: string;
+    };
+  }
+
+  /**
    * Represents the options parameter in the load API
    */
   interface loadOptions {
@@ -75,16 +85,6 @@ declare module "rudder-sdk-js" {
     beaconQueueOptions?: beaconQueueOptions;
     cookieConsentManager?: cookieConsentManager;
     anonymousIdOptions?: anonymousIdOptions;
-  }
-
-  /**
-   * Represents the options parameter for anonymousId
-   */
-  interface anonymousIdOptions {
-    autoCapture: {
-      enabled: boolean;
-      source: string;
-    };
   }
 
   /**
@@ -410,9 +410,7 @@ declare module "rudder-sdk-js" {
   /**
    * To get anonymousId set in the SDK
    */
-  function getAnonymousId(
-    anonymousIdOptions: undefined | anonymousIdOptions
-  ): string;
+  function getAnonymousId(options?: anonymousIdOptions): string;
 
   /**
    * To set anonymousId
@@ -442,6 +440,7 @@ declare module "rudder-sdk-js" {
     queueOptions,
     apiObject,
     apiCallback,
+    anonymousIdOptions,
     load,
     ready,
     reset,
--- a/dist/rudder-sdk-js/index.d.ts
+++ b/dist/rudder-sdk-js/index.d.ts
@@ -74,6 +74,17 @@ declare module "rudder-sdk-js" {
     useBeacon?: boolean; // Defaults to false
     beaconQueueOptions?: beaconQueueOptions;
     cookieConsentManager?: cookieConsentManager;
+    anonymousIdOptions?: anonymousIdOptions;
+  }
+
+  /**
+   * Represents the options parameter for anonymousId
+   */
+  interface anonymousIdOptions {
+    autoCapture: {
+      enabled: boolean;
+      source: string;
+    };
   }
 
   /**
@@ -399,7 +410,9 @@ declare module "rudder-sdk-js" {
   /**
    * To get anonymousId set in the SDK
    */
-  function getAnonymousId(): string;
+  function getAnonymousId(
+    anonymousIdOptions: undefined | anonymousIdOptions
+  ): string;
 
   /**
    * To set anonymousId
--- a/utils/storage/cookie.js
+++ b/utils/storage/cookie.js
@@ -12,7 +12,7 @@ class CookieLocal {
   constructor(options) {
     this._options = {};
     this.options(options);
-    this.isSupportAvailable = this.isSupportAvailable();
+    this.isSupportAvailable = this.checkSupportAvailability();
   }
 
   /**
@@ -74,7 +74,7 @@ class CookieLocal {
    * Function to check cookie support exists or not
    * @returns boolean
    */
-  isSupportAvailable() {
+  checkSupportAvailability() {
     const name = "test_rudder_cookie";
     this.set(name, true);
 
--- a/utils/storage/cookie.js
+++ b/utils/storage/cookie.js
@@ -12,7 +12,7 @@ class CookieLocal {
   constructor(options) {
     this._options = {};
     this.options(options);
-    this.cookieSupported = this.IsCookieSupported();
+    this.isSupportAvailable = this.isSupportAvailable();
   }
 
   /**
@@ -74,11 +74,12 @@ class CookieLocal {
    * Function to check cookie support exists or not
    * @returns boolean
    */
-  IsCookieSupported() {
-    this.set("rudder_cookies", true);
+  isSupportAvailable() {
+    const name = "test_rudder_cookie";
+    this.set(name, true);
 
-    if (this.get("rudder_cookies")) {
-      this.remove("rudder_cookies");
+    if (this.get(name)) {
+      this.remove(name);
       return true;
     }
     return false;
--- a/utils/storage/cookie.js
+++ b/utils/storage/cookie.js
@@ -33,11 +33,9 @@ class CookieLocal {
     });
 
     // try setting a cookie first
-    this.set("test_rudder", true);
-    if (!this.get("test_rudder")) {
+    if (this.IsCookieSupported()) {
       this._options.domain = null;
     }
-    this.remove("test_rudder");
   }
 
   /**
--- a/utils/storage/cookie.js
+++ b/utils/storage/cookie.js
@@ -12,6 +12,7 @@ class CookieLocal {
   constructor(options) {
     this._options = {};
     this.options(options);
+    this.cookieSupported = this.IsCookieSupported();
   }
 
   /**
@@ -31,11 +32,6 @@ class CookieLocal {
       domain,
       samesite: "Lax",
     });
-
-    // try setting a cookie first
-    if (this.IsCookieSupported()) {
-      this._options.domain = null;
-    }
   }
 
   /**
--- a/utils/storage/cookie.js
+++ b/utils/storage/cookie.js
@@ -75,6 +75,20 @@ class CookieLocal {
       return false;
     }
   }
+
+  /**
+   * Function to check cookie support exists or not
+   * @returns boolean
+   */
+  IsCookieSupported() {
+    this.set("rudder_cookies", true);
+
+    if (this.get("rudder_cookies")) {
+      this.remove("rudder_cookies");
+      return true;
+    }
+    return false;
+  }
 }
 
 // Exporting only the instance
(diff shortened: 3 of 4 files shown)


saikumarrs (changes requested):

analytics.js
```diff
@@ -1072,7 +1072,11 @@ class Analytics {
     }
 
     this.eventRepository.initialize(writeKey, serverUrl, options);
-    this.initializeUser();
+    if (options && options.autoAssignAnonId) {
```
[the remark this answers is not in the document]
MoumitaM: Can you think of any future use case as you suggested the below 

```
 enabled: true,
 source: "segment"
```

 fields to be nested in another object
 rather than 

```
 load(<write key>, <data plane url>, {
   autoCaptureAnonymousId: {
      enabled: true,
      source: "segment"
    }
});
```

utils/storage/storage.js
```diff
@@ -248,10 +252,31 @@ class Storage {
...
+  fetchAnonymousId(key) {
+    let anonId;
+    switch (key) {
+      case "segment":
+        if (Store.enabled) {
+          anonId = Store.get(anonymousIdKeyMapper.segment);
+        }
+        if (!anonId) {
+          anonId = Cookie.get(anonymousIdKeyMapper.segment);
+        }
```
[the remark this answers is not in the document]
saikumarrs: Can you please point me to the source of this information?

analytics.js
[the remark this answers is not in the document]
saikumarrs: TBH nothing I can think of but the one I proposed allows extensibility if needed.

analytics.js
```diff
@@ -936,9 +936,9 @@ class Analytics {
     this.storage.clear(flag);
   }
 
-  getAnonymousId() {
+  getAnonymousId(autoAssignAnonId) {
```
saikumarrs, line 939: The TS declaration file updates are missing.

MoumitaM: If we update the TS declaration then it will be publicly available. Anyone can call
rudderanalytics. getAnonymousId("segment");
to fetch segment's anonymous id.

saikumarrs: Because of the change in the signature, it'll be a breaking change for the existing users. Let the users pass the same anonymous ID options object here which would anyway be an optional argument.

utils/storage/storage.js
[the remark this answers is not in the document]
MoumitaM: DM you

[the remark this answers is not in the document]
saikumarrs: Please add the reference information as inline documentation in the code.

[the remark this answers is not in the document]
MoumitaM: Here is the reference [link](https://segment.com/docs/connections/sources/catalog/libraries/website/javascript/#identify).

analytics.js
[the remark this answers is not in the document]

analytics.js
```diff
@@ -1072,7 +1072,11 @@ class Analytics {
     }
 
     this.eventRepository.initialize(writeKey, serverUrl, options);
-    this.initializeUser();
+    if (options && options.autoAssignAnonId) {
+      this.initializeUser(options.autoAssignAnonId);
+    } else {
+      this.initializeUser();
+    }
```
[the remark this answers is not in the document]

utils/storage/storage.js
```diff
@@ -248,10 +252,31 @@ class Storage {
...
 
+  fetchAnonymousId(key) {
+    let anonId;
+    switch (key) {
+      case "segment":
+        if (Store.enabled) {
+          anonId = Store.get(anonymousIdKeyMapper.segment);
+        }
+        if (!anonId) {
+          anonId = Cookie.get(anonymousIdKeyMapper.segment);
```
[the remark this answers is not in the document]
MoumitaM: Addressed

utils/storage/storage.js
```diff
@@ -248,10 +252,31 @@ class Storage {
     );
   }
 
+  fetchAnonymousId(key) {
+    let anonId;
+    switch (key) {
```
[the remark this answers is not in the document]
MoumitaM: Addressed

analytics.js

dist/rudder-sdk-js/index.d.ts
```diff
@@ -74,6 +74,17 @@ declare module "rudder-sdk-js" {
...
+  }
+
+  /**
+   * Represents the options parameter for anonymousId
+   */
+  interface anonymousIdOptions {
+    autoCapture: {
+      enabled: boolean;
+      source: string;
+    };
```
saikumarrs, line 87: Make all of them optional.

dist/rudder-sdk-js/index.d.ts
```diff
@@ -74,6 +74,17 @@ declare module "rudder-sdk-js" {
     useBeacon?: boolean; // Defaults to false
     beaconQueueOptions?: beaconQueueOptions;
     cookieConsentManager?: cookieConsentManager;
+    anonymousIdOptions?: anonymousIdOptions;
+  }
+
+  /**
+   * Represents the options parameter for anonymousId
+   */
+  interface anonymousIdOptions {
```
saikumarrs, line 83: Please define this before it is used.

dist/rudder-sdk-js/index.d.ts
```diff
@@ -399,7 +410,9 @@ declare module "rudder-sdk-js" {
   /**
    * To get anonymousId set in the SDK
    */
-  function getAnonymousId(): string;
+  function getAnonymousId(
+    anonymousIdOptions: undefined | anonymousIdOptions
```
saikumarrs, line 414: The right way to make it an optional argument.
```suggestion
    anonymousIdOptions?: anonymousIdOptions
```

dist/rudder-sdk-js/index.d.ts
saikumarrs, line 83: This should also be exported.

utils/storage/storage.js
```diff
@@ -248,10 +252,45 @@ class Storage {
...
   /**
    * get stored anonymous id
    */
-  getAnonymousId() {
+  getAnonymousId(anonymousIdOptions) {
+    if (
+      anonymousIdOptions &&
+      anonymousIdOptions.autoCapture &&
+      anonymousIdOptions.autoCapture.enabled &&
+      typeof anonymousIdOptions.autoCapture.source === "string"
```
saikumarrs, line 285: Use `get` method instead.

utils/storage/storage.js
```diff
@@ -248,10 +252,45 @@ class Storage {
     );
   }
 
+  fetchAnonymousId(key) {
```
saikumarrs, line 255: Just to make it more readable.
```suggestion
  fetchExternalAnonymousId(key) {
```

utils/storage/storage.js
```diff
@@ -248,10 +252,45 @@ class Storage {
...
-  getAnonymousId() {
+  getAnonymousId(anonymousIdOptions) {
+    if (
+      anonymousIdOptions &&
+      anonymousIdOptions.autoCapture &&
+      anonymousIdOptions.autoCapture.enabled &&
+      typeof anonymousIdOptions.autoCapture.source === "string"
+    ) {
+      const source = anonymousIdOptions.autoCapture.source.toLowerCase();
+      if (Object.keys(anonymousIdKeyMap).includes(source)) {
```
saikumarrs, line 288: These two things can go into the fetch method below.

utils/storage/storage.js
```diff
@@ -248,10 +252,45 @@ class Storage {
...
+    switch (key) {
+      case "segment":
+        if (Store.enabled) {
+          anonId = Store.get(anonymousIdKeyMap[key]);
+        }
+        if (!anonId) {
+          Cookie.set("rudder_cookies", true);
+
+          if (Cookie.get("rudder_cookies")) {
+            Cookie.remove("rudder_cookies");
```
saikumarrs, line 266: Refactor this into a separate method.

dist/rudder-sdk-js/index.d.ts
MoumitaM: Done

dist/rudder-sdk-js/index.d.ts
MoumitaM: Done

dist/rudder-sdk-js/index.d.ts
MoumitaM: Done

dist/rudder-sdk-js/index.d.ts
MoumitaM: Done

utils/storage/storage.js
MoumitaM: Done

utils/storage/storage.js
MoumitaM: Done

utils/storage/storage.js
MoumitaM: Done

utils/storage/storage.js
saikumarrs: The object structure is incorrectly used.

utils/storage/storage.js
```diff
@@ -248,10 +251,42 @@ class Storage {
...
+  }
+
   /**
    * get stored anonymous id
    */
-  getAnonymousId() {
+  getAnonymousId(anonymousIdOptions) {
+    const source = get(anonymousIdOptions, "source");
+    if (get(anonymousIdOptions, "enabled") && typeof source === "string") {
+      const anonId = this.fetchExternalAnonymousId(source.toLowerCase());
```
saikumarrs, line 286: Let's leave the `toLowerCase` conversion to the fetch method.

utils/storage/cookie.js
```diff
@@ -75,6 +75,20 @@ class CookieLocal {
       return false;
     }
   }
+
+  /**
+   * Function to check cookie support exists or not
+   * @returns boolean
+   */
+  IsCookieSupported() {
```
saikumarrs, line 83: Can you re-use this method for the above cookie check?

MoumitaM: done

utils/storage/storage.js
MoumitaM: done

utils/storage/storage.js
MoumitaM: If we use `get` method in case the customer set 

`anonymousIdOptions: "random string"`

`get(anonymousIdOptions, "autoCapture.enabled") `
and
`get(anonymousIdOptions, "autoCapture.source")
`
is returning  "random string".

saikumarrs: That's an instrumentation problem and anyway it leads to incorrect results.

utils/storage/storage.js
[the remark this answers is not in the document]
saikumarrs: Gentle reminder to address this.

utils/storage/storage.js
MoumitaM: Can you elaborate on 

> The object structure is incorrectly used.
> it leads to incorrect results.

I have tested it properly and the results don’t seem to be incorrect.

utils/storage/storage.js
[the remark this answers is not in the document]
MoumitaM: Addressed

utils/storage/storage.js
saikumarrs: The comment was for the previous 'non-get' implementation.

saikumarrs: I was referring to this commit,
https://github.com/rudderlabs/rudder-sdk-js/pull/497/commits/e2b3ca105ed158e4ac0a1360971788d42c688abd

where `autoCapture` object is missed while accessing the options.

MoumitaM: so are we good with this approach?

saikumarrs: No. Let's use the `get` method to access multi-level fields.

MoumitaM: Addressed

utils/storage/storage.js
```diff
@@ -248,10 +251,49 @@ class Storage {
...
+  }
+
   /**
    * get stored anonymous id
    */
-  getAnonymousId() {
+  getAnonymousId(anonymousIdOptions) {
+    const source = get(anonymousIdOptions, "autoCapture.source");
+    if (
+      get(anonymousIdOptions, "autoCapture.enabled") &&
```
saikumarrs, line 290:
```suggestion
      get(anonymousIdOptions, "autoCapture.enabled") === true &&
```

utils/storage/storage.js
```diff
@@ -248,10 +251,49 @@ class Storage {
...
+    if (
+      get(anonymousIdOptions, "autoCapture.enabled") === true &&
+      typeof source === "string"
+    ) {
+      const anonId = this.fetchExternalAnonymousId(source);
+      if (anonId) return anonId;
+    }
+
     return this.parse(
       this.decryptValue(this.storage.get(defaults.user_storage_anonymousId))
```
saikumarrs, line 298: We should first check if the anonymous ID is not available in RS storage and if it is unavailable, then try to retrieve it from the auto-capture source.

MoumitaM: Addressed

saikumarrs (approved):

pallabmaiti (dismissed):

saikumarrs:
PR description doesn't match with the implementation.

utils/storage/cookie.js
```diff
@@ -12,6 +12,7 @@ class CookieLocal {
   constructor(options) {
     this._options = {};
     this.options(options);
+    this.cookieSupported = this.IsCookieSupported();
```
saikumarrs, line 15: The class already has 'cookie' in its name. Let's rename the variable to `isSupportAvailable`.

utils/storage/cookie.js
```diff
@@ -75,6 +69,20 @@ class CookieLocal {
...
+   * @returns boolean
+   */
+  IsCookieSupported() {
+    this.set("rudder_cookies", true);
+
+    if (this.get("rudder_cookies")) {
+      this.remove("rudder_cookies");
+      return true;
+    }
+    return false;
```
saikumarrs, line 84: Two things:
- Rename it to `test_rudder_cookie`
- Reuse the cookie name rather than repeat it in multiple places.

utils/storage/storage.js
```diff
@@ -248,13 +251,81 @@ class Storage {
...
 
+  /**
+   * Function to fetch anonymousId from external source
+   * @param {string} key source of the anonymousId
+   * @returns string
+   */
+  fetchExternalAnonymousId(source) {
+    let anonId;
+    const key = source.toLowerCase();
+    if (!Object.keys(anonymousIdKeyMap).includes(key)) {
```
saikumarrs, line 262: Let's add a rationale for this function's code flow.

utils/storage/storage.js
```diff
@@ -248,13 +251,81 @@ class Storage {
...
    */
-  getAnonymousId() {
-    return this.parse(
+  getAnonymousId(anonymousIdOptions) {
+    // fetch the rl_anonymous_id from storage
+    const rlAnonymousId = this.parse(
       this.decryptValue(this.storage.get(defaults.user_storage_anonymousId))
     );
+    if (rlAnonymousId) {
+      return rlAnonymousId; // return if already present
```
saikumarrs, line 314: Let's add a rationale for this function's code flow.

utils/storage/cookie.js
saikumarrs: Same with the method name.

utils/storage/storage.js
MoumitaM: You mean the function or the if block?

saikumarrs: The function.

MoumitaM: Already added use cases what else could be added?

utils/storage/storage.js
```diff
@@ -248,13 +251,84 @@ class Storage {
...
-  getAnonymousId() {
-    return this.parse(
+  getAnonymousId(anonymousIdOptions) {
+    // fetch the rl_anonymous_id from storage
+    const rlAnonymousId = this.parse(
       this.decryptValue(this.storage.get(defaults.user_storage_anonymousId))
     );
+    /**
+     * If rl_anonymous_id is already present return the value rather then executing the rest of the block
+     */
```
saikumarrs, line 315:
```suggestion
    // If RS's anonymous ID is already present, return from here.
    // The user, while migrating from a different analytics SDK, will only need to auto-capture the anonymous ID when the RS SDK loads for the first time. 
    // The captured anonymous ID would be available in RS's persistent storage for all the subsequent SDK runs. 
    // So, instead of always grabbing the ID from the migration source when the options are specified, it is first checked in the RS's persistent storage.
    // Moreover, the user can also clear the anonymous ID from the storage via the 'reset' API, which renders the migration source's data useless.
    ```

File: utils/storage/storage.js (hunk 21 above)
Comment (saikumarrs) [reply utils/storage/storage.js]: OBE with this suggestion,
https://github.com/rudderlabs/rudder-sdk-js/pull/497/files#r856312675.
Hence, resolving this conversation.

File: utils/storage/cookie.js (hunk 18 above)
Comment (MoumitaM) [reply utils/storage/cookie.js]: Addressed

File: utils/storage/cookie.js (hunk 19 above)
Comment (MoumitaM) [reply utils/storage/cookie.js]: Addressed

File: utils/storage/cookie.js [hunk 23]
```diff
@@ -12,6 +12,7 @@ class CookieLocal {
   constructor(options) {
     this._options = {};
     this.options(options);
+    this.isSupportAvailable = this.isSupportAvailable();
```
saikumarrs, line 15: It's a bad idea to name both of them identical.

utils/storage/storage.js
```diff
@@ -248,13 +251,87 @@ class Storage {
...
+    // fetch the rl_anonymous_id from storage
+    const rlAnonymousId = this.parse(
       this.decryptValue(this.storage.get(defaults.user_storage_anonymousId))
     );
+    // If RS's anonymous ID is available, return from here.
+    // The user, while migrating from a different analytics SDK, will only need to auto-capture the anonymous ID when the RS SDK loads for the first time. 
+    // The captured anonymous ID would be available in RS's persistent storage for all the subsequent SDK runs. 
+    // So, instead of always grabbing the ID from the migration source when the options are specified, it is first checked in the RS's persistent storage.
+    // Moreover, the user can also clear the anonymous ID from the storage via the 'reset' API, which renders the migration source's data useless.
+    ```
```
saikumarrs, line 318: Seems like a typo.

utils/storage/storage.js
```diff
@@ -248,13 +251,87 @@ class Storage {
...
+   *
    */
-  getAnonymousId() {
-    return this.parse(
+  getAnonymousId(anonymousIdOptions) {
+    // fetch the rl_anonymous_id from storage
+    const rlAnonymousId = this.parse(
       this.decryptValue(this.storage.get(defaults.user_storage_anonymousId))
     );
+    // If RS's anonymous ID is available, return from here.
```
saikumarrs, line 313: The statements are lengthy. Can we make them multilined?

utils/storage/cookie.js
MoumitaM: Addressed

utils/storage/storage.js
MoumitaM: Addressed

utils/storage/storage.js
MoumitaM: Addressed

saikumarrs (approved):

PR #507, opened 2022-04-25
Feature: Allow API calls before load

MoumitaM:
## Description of the change

> This feature will enable the SDK methods to be called before load.
- All the previously called methods will be pushed to the call stack after load in the same order.
- Ex: Say the call stack currently have [page, identify, load, track]. It will become [load, page, identify, track]

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/507)

### DIFF
--- a/analytics.js
+++ b/analytics.js
@@ -1336,10 +1336,13 @@ const eventsPushedAlready =
   !!window.rudderanalytics &&
   window.rudderanalytics.push == Array.prototype.push;
 
-const argumentsArray = window.rudderanalytics;
+let argumentsArray = window.rudderanalytics;
+
+const apiCallsBeforeLoad = [];
 
 while (argumentsArray && argumentsArray[0] && argumentsArray[0][0] !== "load") {
-  argumentsArray.shift();
+  // Collect the methods that are called before load in "apiCallsBeforeLoad" array
+  apiCallsBeforeLoad.push(argumentsArray.shift());
 }
 if (
   argumentsArray &&
@@ -1351,6 +1354,14 @@ if (
   logger.debug("=====from init, calling method:: ", method);
   instance[method](...argumentsArray[0]);
   argumentsArray.shift();
+  /**
+   * After load is called
+   * Check if the array has any method that is called before load
+   * If present, Push the calls at the front of the call stack
+   */
+  if (apiCallsBeforeLoad.length) {
+    argumentsArray = apiCallsBeforeLoad.concat(argumentsArray);
+  }
 }
 
 // once loaded, parse querystring of the page url to send events
--- a/analytics.js
+++ b/analytics.js
@@ -1336,7 +1336,7 @@ const eventsPushedAlready =
   !!window.rudderanalytics &&
   window.rudderanalytics.push == Array.prototype.push;
 
-const argumentsArray = window.rudderanalytics;
+const argumentsArray = window.rudderanalytics || [];
 /**
  * Iterate the call stack until we find load call and
  * then shift it to the beginning.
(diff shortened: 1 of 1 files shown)


analytics.js
```diff
@@ -1336,10 +1336,17 @@ const eventsPushedAlready =
...
+/**
+ * Usage of the below variable:
+ * It will be used to store the methods that are called before load.
+ */
+const apiCallsBeforeLoad = [];
 
 while (argumentsArray && argumentsArray[0] && argumentsArray[0][0] !== "load") {
-  argumentsArray.shift();
+  // Collect the methods that are called before load in "apiCallsBeforeLoad" array
+  apiCallsBeforeLoad.push(argumentsArray.shift());
```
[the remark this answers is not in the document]

analytics.js
```diff
@@ -1337,10 +1337,20 @@ const eventsPushedAlready =
...
+ * then shift it to the beginning.
+ *
+ * Ex: Say the call stack currently have [page, identify, load, track]
+ * It will become [load, page, identify, track]
+ */
+for (let i = 0; i < argumentsArray.length; i++) {
+  if (argumentsArray[i] && argumentsArray[i][0] === "load") {
+    argumentsArray.unshift(argumentsArray.splice(i, 1)[0]);
+    break;
+  }
```
saikumarrs, line 1351: If no `load` call was found, others calls should not be processed.

analytics.js
```diff
@@ -1337,10 +1337,20 @@ const eventsPushedAlready =
...
-while (argumentsArray && argumentsArray[0] && argumentsArray[0][0] !== "load") {
-  argumentsArray.shift();
+/**
+ * Iterate the call stack until we find load call and
+ * then shift it to the beginning.
+ *
+ * Ex: Say the call stack currently have [page, identify, load, track]
+ * It will become [load, page, identify, track]
+ */
+for (let i = 0; i < argumentsArray.length; i++) {
```
saikumarrs, line 1347: Use a while loop here as this array can be simultaneously updated by the client code.

saikumarrs (changes requested):
Added requested changes as comments.

analytics.js
MoumitaM: There might be scenarios where the load method is called on button click/ called from the body, not from the head tag. For such cases load won’t be present in argumentsArray. We should let other calls to be processed.

analytics.js
```diff
@@ -1337,10 +1337,30 @@ const eventsPushedAlready =
...
+    break;
+  }
+  i += 1;
 }
+/**
+ * When load is not present in the call stack restrict other events from proceeding
+ */
+if (!loadPresent) {
+  argumentsArray.length = 0;
+}
```
saikumarrs, line 1362: There is no need for this check. Instead, you can add an else statement on Ln 1374.

analytics.js
```diff
@@ -1336,21 +1336,34 @@ const eventsPushedAlready =
   !!window.rudderanalytics &&
   window.rudderanalytics.push == Array.prototype.push;
 
-const argumentsArray = window.rudderanalytics;
-
-while (argumentsArray && argumentsArray[0] && argumentsArray[0][0] !== "load") {
-  argumentsArray.shift();
+const argumentsArray = window.rudderanalytics || [];
```
saikumarrs, line 1339:
```suggestion
const argumentsArray = window.rudderanalytics;
```

MoumitaM: If someone only loads the CDN url, there will be no `argumentsArray`. In that case from Ln 1348 and 1366 they will get error `length of undefined`.

analytics.js
```diff
@@ -1336,21 +1336,34 @@ const eventsPushedAlready =
...
   argumentsArray[0].shift();
   logger.debug("=====from init, calling method:: ", method);
   instance[method](...argumentsArray[0]);
   argumentsArray.shift();
+} else {
+  /**
+   * When load is not present in the call stack restrict other events from proceeding
+   */
+  argumentsArray.length = 0;
 }
```
saikumarrs, line 1367: I redact my previous statement. Let's not remove the elements from the array in this scenario as I realized the situation would be the same for the query string API where they'd be waiting for the `load` API call.

analytics.js

MoumitaM: It wasn’t there in the previous implementation. We added that because in the case of undefined argumentsArray users will get an error. 
The undefined case is handled and if the users use our loading code snippet as mentioned in our docs there will not be any case of `argumentsArray` being anything else other than an array. 
Let’s not add the extra checking that might lead to some other issue that at this moment we can not think of.

analytics.js
```diff
@@ -1337,12 +1337,24 @@ const eventsPushedAlready =
...
+  if (argumentsArray[i] && argumentsArray[i][0] === "load") {
+    argumentsArray.unshift(argumentsArray.splice(i, 1)[0]);
+    break;
+  }
+  i += 1;
 }
+
 if (
-  argumentsArray &&
+  Array.isArray(argumentsArray) &&
```
saikumarrs, line 1357: Let's not repeat this check.

saikumarrs (approved):

PR #582, opened 2022-07-25
Fix: source config fetch error handled

MoumitaM:
## Description of the change

> Added error handling in case of source config fetch fails or returns an empty/no response.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/582)

### DIFF
--- a/analytics.js
+++ b/analytics.js
@@ -174,9 +174,19 @@ class Analytics {
         return;
       }
 
-      if (typeof response === "string") {
-        response = JSON.parse(response);
-      }
+var response = responseVal;
+try {
+  if (typeof responseVal === "string") {
+    response = JSON.parse(responseVal);
+  }
+  
+  // Do not proceed if the ultimate response value is not an object
+  if (!response || typeof response !== "object" || Array.isArray(response)) {
+    throw new Error("Invalid source configuration");
+  }
+} catch (err) {
+  handleError(err);
+}
 
       // Fetch Error reporting enable option from sourceConfig
       const isErrorReportEnabled = get(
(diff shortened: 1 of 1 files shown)


analytics.js
```diff
@@ -169,6 +169,11 @@ class Analytics {
...
     try {
       logger.debug(`===in process response=== ${status}`);
+      if (!response) {
+        logger.debug(`No config found`);
+        return;
+      }
+
       if (typeof response === "string") {
         response = JSON.parse(response);
       }
```
saikumarrs, line 179: You can move the validation to this block by directly validating the JSON response. Ensure to add error handling as well.

MoumitaM: In case of error while fetching the source config it calls `processResponse` only with status parameter.
`if (typeof response === "string") {
        response = JSON.parse(response);
      }`
  Now the response is `undefined` and it doesn’t enter this block. So if I move the validation inside it won’t be effective .

saikumarrs: Try changing the logic.

```suggestion
var response = responseVal;
try {
  if (typeof responseVal === "string") {
    response = JSON.parse(responseVal);
  }
  
  // Do not proceed if the ultimate response value is not an object
  if (!response || typeof response !== "object" || Array.isArray(response)) {
    throw new Error("Invalid source configuration");
  }
} catch (err) {
  handleError(err);
}
```

MoumitaM: What is the benefit of having an extra variable like `responseVal`? We can use the `response`.

saikumarrs: We should not modify the incoming arguments. It's an error as per eslint as well.

saikumarrs (approved):

PR #587, opened 2022-07-27
Filter script loading errors from native SDKs

MoumitaM:
## Description of the change
- Add a new attribute `isNonNativeSDK`. Default is `false` for all the native SDKs loaded by RS.
- Added checking in util to filter dataset with isNative set to true.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/587)


utsabc:
I might not have enough context but does all native 3rd party sdk support this flag?

MoumitaM:
We are intentionally adding this flag to filter based on that.
You can add any field to the dataset. [Ref](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset)

utsabc:
> We are intentionally adding this flag to filter based on that. You can add any field to the dataset. [Ref](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset)

But How we will use this to filter loading of  third party sdks?

utils/utils.js
```diff
@@ -137,6 +137,9 @@ function handleError(error, analyticsInstance) {
       // Discard errors of scripts that are not loaded by the SDK
       if (error.target.dataset && error.target.dataset.loader !== LOAD_ORIGIN)
         return;
+      // Discard script loading errors generated from native SDKs
+      if (error.target.dataset && error.target.dataset.isNative === IS_NATIVE_SCRIPT)
+        return;
```
saikumarrs, line 142:
```suggestion
      // Discard all script errors unless they are of integration SDK scripts
      if (error.target.dataset && (error.target.dataset.loader !== LOAD_ORIGIN || error.target.dataset.isNative === IS_NATIVE_SCRIPT))
        return;
```

integrations/ScriptLoader.js
```diff
@@ -4,6 +4,7 @@ import logger from "../utils/logUtil";
 const defaultAsyncState = true;
 
 export const LOAD_ORIGIN = "RS_JS_SDK";
+export const IS_NATIVE_SCRIPT = "true";
```
saikumarrs, line 7: There is no need to make this a constant.

integrations/ScriptLoader.js
```diff
@@ -13,6 +14,7 @@ const ScriptLoader = (id, src, async = defaultAsyncState) => {
   js.type = "text/javascript";
   js.id = id;
   js.dataset.loader = LOAD_ORIGIN;
+  js.dataset.isNative = IS_NATIVE_SCRIPT;
```
saikumarrs, line 17: Why are you setting this by default? Now, any script loaded via this utility will become a native script.

integrations/ScriptLoader.js
```diff
@@ -4,6 +4,7 @@ import logger from "../utils/logUtil";
 const defaultAsyncState = true;
 
 export const LOAD_ORIGIN = "RS_JS_SDK";
+export const IS_NATIVE_SCRIPT = "true";
 
 const ScriptLoader = (id, src, async = defaultAsyncState) => {
```
saikumarrs, line 9: Change this method signature to convert the third argument into an options object.
Also, change `isNative` to use `isRudderSDK` (invert the definition). Default `isRudderSDK` to true as it will minimize the overall changes needed and also prevents any developer errors going forward.


```
options: {
  async: true, // fallback to `defaultAsyncState` if not sent
  isRudderSDK: true // default to true if not sent
}
```

saikumarrs (changes requested):

integrations/ScriptLoader.js
MoumitaM: In the v1 version, we only load native scripts.
For v1.1 I will add logic here.

saikumarrs: As an example to counter that argument, we load polyfill script in both versions. Hence, let's keep the implementation the same across.

integrations/ScriptLoader.js
MoumitaM: Updated the filtering logic.

saikumarrs: Convert the arguments after `id` and `src` to `options` like I suggested previously.

integrations/ScriptLoader.js
```diff
@@ -4,15 +4,17 @@ import logger from "../utils/logUtil";
...
-const ScriptLoader = (id, src, async = defaultAsyncState) => {
+const ScriptLoader = (id, src, async = defaultAsyncState, isNonNativeSDK = defaultNonNativeSDKState) => {
   logger.debug(`in script loader=== ${id}`);
   const js = document.createElement("script");
   js.src = src;
   js.async = async === undefined ? defaultAsyncState : async;
   js.type = "text/javascript";
   js.id = id;
   js.dataset.loader = LOAD_ORIGIN;
+  js.dataset.isNonNativeSDK = isNonNativeSDK;
```
saikumarrs, line 17: Set this field only if the user provides the value. No default.

utils/utils.js
```diff
@@ -134,8 +134,8 @@ function handleError(error, analyticsInstance) {
       // Discard all the non-script loading errors
       if (error.target && error.target.localName !== "script") return;
 
-      // Discard errors of scripts that are not loaded by the SDK
-      if (error.target.dataset && error.target.dataset.loader !== LOAD_ORIGIN)
+      // Discard all script errors unless they are of integration SDK scripts
```
saikumarrs, line 137:
```suggestion
      // Discard script errors that are not originated at SDK or from native SDKs
```

utils/utils.js
```diff
@@ -134,8 +134,8 @@ function handleError(error, analyticsInstance) {
...
+      if (error.target.dataset && (error.target.dataset.loader !== LOAD_ORIGIN || error.target.dataset.isNonNativeSDK === defaultNonNativeSDKState))
```
saikumarrs, line 138:
```suggestion
      if (error.target.dataset && (error.target.dataset.loader !== LOAD_ORIGIN || error.target.dataset.isNonNativeSDK !== true))
```

saikumarrs (changes requested):

utils/utils.js
MoumitaM: element.dataset.property converts the value to string.

integrations/ScriptLoader.js
MoumitaM: Addressed.

integrations/ScriptLoader.js
MoumitaM: Addressed

utils/utils.js
MoumitaM: Addressed

integrations/ScriptLoader.js
```diff
@@ -2,17 +2,28 @@
...
+) => {
   logger.debug(`in script loader=== ${id}`);
   const js = document.createElement("script");
   js.src = src;
-  js.async = async === undefined ? defaultAsyncState : async;
+  js.async = options.async === undefined ? defaultAsyncState : options.async;
   js.type = "text/javascript";
   js.id = id;
   js.dataset.loader = LOAD_ORIGIN;
+  if (options.isNonNativeSDK) {
```
saikumarrs, line 24:
```suggestion
  if (options.isNonNativeSDK !== undefined) {
```

integrations/ScriptLoader.js
```diff
@@ -2,17 +2,28 @@
 import logger from "../utils/logUtil";
 
 const defaultAsyncState = true;
+const defaultNonNativeSDKState = "false";
 
 export const LOAD_ORIGIN = "RS_JS_SDK";
 
-const ScriptLoader = (id, src, async = defaultAsyncState) => {
+const ScriptLoader = (
```
saikumarrs, line 9: There are references to the ScriptLoader method in the integrations that need updates (Ex: GoogleOptimize). Check all of them thoroughly.

integrations/ScriptLoader.js
```diff
@@ -2,17 +2,28 @@
 import logger from "../utils/logUtil";
 
 const defaultAsyncState = true;
+const defaultNonNativeSDKState = "false";
```
saikumarrs, line 5: We don't need these anymore.

integrations/ScriptLoader.js
```diff
@@ -2,17 +2,28 @@
...
 export const LOAD_ORIGIN = "RS_JS_SDK";
 
-const ScriptLoader = (id, src, async = defaultAsyncState) => {
+const ScriptLoader = (
+  id,
+  src,
+  options = {
+    async: defaultAsyncState,
+    isNonNativeSDK: defaultNonNativeSDKState,
+  }
```
saikumarrs, line 15: We can get rid of these.

saikumarrs: Defaults.

integrations/ScriptLoader.js
MoumitaM: Addressed.

integrations/ScriptLoader.js
MoumitaM: Addressed.

MoumitaM: Only removed `const defaultNonNativeSDKState = "false";` this line

integrations/ScriptLoader.js
MoumitaM: Addressed

integrations/ScriptLoader.js
MoumitaM: Addressed

integrations/ScriptLoader.js
MoumitaM: Addressed

saikumarrs (dismissed):

MoumitaM:
@utsabc 
> But How we will use this to filter loading of third party sdks?

We will filter third-party loading errors not the native SDKs.

saikumarrs (approved):

utsabc (approved):

PR #590, opened 2022-07-29
feature/mouseflow: new Integration Onboarding

sauravlal-rudderstack:
## Description of the change

> Mouseflow - new Integration for web device mode.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

---
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/590)

### DIFF
--- a/integrations/Mouseflow/browser.js
+++ b/integrations/Mouseflow/browser.js
@@ -4,6 +4,7 @@ import get from "get-value";
 import { NAME } from "./constants";
 import logger from "../../utils/logUtil";
 import ScriptLoader from "../ScriptLoader";
+import setCustomVariables from "./utils";
 import { isDefinedAndNotNull } from "../utils/commonUtils";
 
 class Mouseflow {
@@ -41,35 +42,37 @@ class Mouseflow {
     if (integrations && integrations[NAME]) {
       const tags = integrations[NAME];
       if (isDefinedAndNotNull(tags)) {
-        Object.entries(tags).forEach((item) => {
-          const [key, value] = item;
-          if (typeof value === "string")
-            window._mfq.push(["setVariable", key, value]);
-        });
+        setCustomVariables(tags);
       }
     }
   }
 
   /**
    * Identify.
+   * for supporting userId or email
    * Ref: https://js-api-docs.mouseflow.com/#identifying-a-user
+   * for supporting user traits and tags
+   * Ref: https://js-api-docs.mouseflow.com/#setting-a-custom-variable
    * @param {Identify} identify
    */
   identify(rudderElement) {
     const { message } = rudderElement;
+    const { traits } = message.context;
     const email =
       get(message, "context.traits.email") || get(message, "traits.email");
     const userId = message.userId || email;
     window._mfq.push(["stop"]);
     if (userId) window.mouseflow.identify(userId);
     window.mouseflow.start();
+    setCustomVariables(traits);
+    this.addTags(message);
   }
 
   /**
    * Track - tracks an event for a specific user
-   * Set custom tags
+   * for supporting event
    * Ref: https://js-api-docs.mouseflow.com/#tagging-a-recording
-   * Set custom Variables
+   * for supporting properties and tags
    * Ref: https://js-api-docs.mouseflow.com/#setting-a-custom-variable
    * @param {Track} track
    */
@@ -83,16 +86,13 @@ class Mouseflow {
       return;
     }
     window._mfq.push(["tag", event]);
-    Object.entries(properties).forEach((item) => {
-      const [key, value] = item;
-      if (typeof value === "string")
-        window._mfq.push(["setVariable", key, value]);
-    });
+    setCustomVariables(properties);
     this.addTags(message);
   }
 
   /**
    * Page.
+   * for supporting path of Page
    * Ref: https://js-api-docs.mouseflow.com/#setting-a-virtual-path
    * @param {Page} page
    */
--- a/integrations/Mouseflow/browser.js
+++ b/integrations/Mouseflow/browser.js
@@ -1,0 +1,105 @@
+/* eslint-disable no-underscore-dangle */
+/* eslint-disable class-methods-use-this */
+import get from "get-value";
+import { NAME } from "./constants";
+import logger from "../../utils/logUtil";
+import ScriptLoader from "../ScriptLoader";
+
+class Mouseflow {
+  constructor(config) {
+    this.websiteId = config.websiteId;
+    this.name = NAME;
+  }
+
+  init() {
+    logger.debug("===In init mouseflow===");
+    window._mfq = window._mfq || [];
+    // ScriptLoader(
+    //   "mouseflow-integration",
+    //   `https://cdn.mouseflow.com/projects/${this.websiteId}.js`
+    // );
+    (function () {
+      var mf = document.createElement("script");
+      mf.type = "text/javascript";
+      mf.defer = true;
+      mf.src = `//cdn.mouseflow.com/projects/${this.websiteId}.js`;
+      document.getElementsByTagName("head")[0].appendChild(mf);
+    })();
+  }
+
+  isLoaded() {
+    logger.debug("===In isLoaded mouseflow===");
+    return !!window.mouseflow && typeof window.mouseflow === "object";
+  }
+
+  isReady() {
+    logger.debug("===In isReady mouseflow===");
+    return !!window._mfq;
+  }
+
+  /**
+   * Identify.
+   * Ref: https://js-api-docs.mouseflow.com/#identifying-a-user
+   * @param {Identify} identify
+   */
+  identify(rudderElement) {
+    const { message } = rudderElement;
+    const email =
+      get(message, "context.traits.email") || get(message, "traits.email");
+    const userId = message.userId || email;
+    _mfq.push(["stop"]);
+    if (userId) window.mouseflow.identify(userId);
+    mouseflow.start();
+  }
+
+  /**
+   * Track - tracks an event for a specific user
+   * Set custom tags
+   * Ref: https://js-api-docs.mouseflow.com/#tagging-a-recording
+   * Set custom Variables
+   * Ref: https://js-api-docs.mouseflow.com/#setting-a-custom-variable
+   * @param {Track} track
+   */
+  track(rudderElement) {
+    logger.debug("=== In mouseflow track ===");
+
+    const { message } = rudderElement;
+    const { event, properties } = message;
+    if (!event) {
+      logger.error("[mouseflow]: Event name from track call is missing!!===");
+      return;
+    }
+    window._mfq.push(["tag", event]);
+    Object.entries(properties).forEach((_ref) => {
+      const [_key, _value] = _ref;
+      window._mfq.push(["setVariable", _key, _value]);
+    });
+  }
+
+  /**
+   * Page.
+   * Ref: https://js-api-docs.mouseflow.com/#setting-a-virtual-path
+   * @param {Page} page
+   */
+  page(rudderElement) {
+    logger.debug("=== In mouseflow Page ===");
+    const { name, category, properties } = rudderElement.message;
+    let eventName;
+    if (!name && !category) {
+      eventName = `Viewed Page`;
+    } else if (!name && category) {
+      eventName = `Viewed ${category} Page`;
+    } else if (name && !category) {
+      eventName = `Viewed ${name} Page`;
+    } else {
+      eventName = `Viewed ${category} ${name} Page`;
+    }
+    window._mfq.push(["newPageView", eventName]);
+    Object.entries(properties).forEach((_ref) => {
+      const [_key, _value] = _ref;
+      window._mfq.push(["setVariable", _key, _value]);
+    });
+  }
+}
+
+export default Mouseflow;
--- a/integrations/Mouseflow/browser.js
+++ b/integrations/Mouseflow/browser.js
@@ -38,12 +38,9 @@ class Mouseflow {
    * Ref: https://js-api-docs.mouseflow.com/#setting-a-custom-variable
    */
   addTags(message) {
-    const { integrations } = message;
-    if (integrations && integrations[NAME]) {
-      const tags = integrations[NAME];
-      if (isDefinedAndNotNull(tags)) {
-        setCustomVariables(tags);
-      }
+    const tags = get(message, `integrations.${NAME}.tags`);
+    if (isDefinedAndNotNull(tags)) {
+      setCustomVariables(tags);
     }
   }
 
@@ -62,7 +59,7 @@ class Mouseflow {
       get(message, "context.traits.email") || get(message, "traits.email");
     const userId = message.userId || email;
     window._mfq.push(["stop"]);
-    if (userId) window.mouseflow.identify(userId);
+    if (userId) window._mfq.push(["identify", userId]);
     window.mouseflow.start();
     setCustomVariables(traits);
     this.addTags(message);
@@ -99,7 +96,7 @@ class Mouseflow {
   page(rudderElement) {
     logger.debug("=== In mouseflow Page ===");
     const tabPath =
-      rudderElement.message.properties?.path ||
+      rudderElement.message.properties.path ||
       rudderElement.message.context.path;
     if (tabPath) window._mfq.push(["newPageView", tabPath]);
   }
--- a/integrations/Mouseflow/browser.js
+++ b/integrations/Mouseflow/browser.js
@@ -1,11 +1,9 @@
 /* eslint-disable no-underscore-dangle */
 /* eslint-disable class-methods-use-this */
-import get from "get-value";
 import { NAME } from "./constants";
 import logger from "../../utils/logUtil";
 import ScriptLoader from "../ScriptLoader";
-import setCustomVariables from "./utils";
-import { isDefinedAndNotNull } from "../utils/commonUtils";
+import { setCustomVariables, addTags } from "./utils";
 
 class Mouseflow {
   constructor(config) {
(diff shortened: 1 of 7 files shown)


integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,100 @@
...
+    if (!event) {
+      logger.error("[mouseflow]: Event name from track call is missing!!===");
+      return;
+    }
+    window._mfq.push(["tag", event]);
+    Object.entries(properties).forEach((_ref) => {
+      const [_key, _value] = _ref;
+      window._mfq.push(["setVariable", _key, _value]);
+    });
+  }
```
ItsSudip, line 72: We should also support tags from integrations object similar to `vero`.

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,100 @@
...
+    if (!name && !category) {
+      eventName = `Viewed Page`;
+    } else if (!name && category) {
+      eventName = `Viewed ${category} Page`;
+    } else if (name && !category) {
+      eventName = `Viewed ${name} Page`;
+    } else {
+      eventName = `Viewed ${category} ${name} Page`;
+    }
+    window._mfq.push(["newPageView", eventName]);
```
ItsSudip, line 92: We shouldn't send a string like that. It should be the path.

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,100 @@
...
+    Object.entries(properties).forEach((_ref) => {
+      const [_key, _value] = _ref;
+      window._mfq.push(["setVariable", _key, _value]);
+    });
```
ItsSudip, line 96: No need to add properties as of now.

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,100 @@
...
+    const { event, properties } = message;
+    if (!event) {
+      logger.error("[mouseflow]: Event name from track call is missing!!===");
+      return;
+    }
+    window._mfq.push(["tag", event]);
+    Object.entries(properties).forEach((_ref) => {
+      const [_key, _value] = _ref;
+      window._mfq.push(["setVariable", _key, _value]);
+    });
```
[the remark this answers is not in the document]
sauravlal-rudderstack: Updated

integrations/Mouseflow/browser.js
sauravlal-rudderstack: Yeah, done!

integrations/Mouseflow/browser.js
sauravlal-rudderstack: yeah updated

integrations/Mouseflow/browser.js
sauravlal-rudderstack: yeah

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,100 @@
...
+  init() {
+    logger.debug("===In init mouseflow===");
+    window._mfq = window._mfq || [];
+    (function (websiteId) {
+      var mf = document.createElement("script");
+      mf.type = "text/javascript";
+      mf.defer = true;
+      mf.src = `https://cdn.mouseflow.com/projects/${websiteId}.js`;
+      document.getElementsByTagName("head")[0].appendChild(mf);
+    })(this.websiteId);
```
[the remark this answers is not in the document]
sauravlal-rudderstack: yes, done!

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,108 @@
...
+    if (integrations && integrations[NAME]) {
+      const tags = integrations[NAME];
+      if (isDefinedAndNotNull(tags)) {
+        Object.entries(tags).forEach((item) => {
+          const [key, value] = item;
+          if (typeof value === "string")
+            window._mfq.push(["setVariable", key, value]);
+        });
+      }
+    }
```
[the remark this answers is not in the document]
sauravlal-rudderstack: updated

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,108 @@
...
+   * Ref: https://js-api-docs.mouseflow.com/#identifying-a-user
+   * @param {Identify} identify
+   */
+  identify(rudderElement) {
+    const { message } = rudderElement;
+    const email =
+      get(message, "context.traits.email") || get(message, "traits.email");
+    const userId = message.userId || email;
+    window._mfq.push(["stop"]);
+    if (userId) window.mouseflow.identify(userId);
```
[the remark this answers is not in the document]
sauravlal-rudderstack: yeah, updated

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,108 @@
...
+    logger.debug("===In isReady mouseflow===");
+    return !!window._mfq;
+  }
+
+  /*
+   * Add tags
+   * Set custom Variables
+   * Ref: https://js-api-docs.mouseflow.com/#setting-a-custom-variable
+   */
+  addTags(message) {
```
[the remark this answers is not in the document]
sauravlal-rudderstack: We are iterating each key-value pair and if the value is object, we'll push it as custom variables for mouseflow.
Should we support the tag??
@saikumarrs 

CC: @ItsSudip

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,108 @@
+/* eslint-disable no-underscore-dangle */
+/* eslint-disable class-methods-use-this */
+import get from "get-value";
+import { NAME } from "./constants";
+import logger from "../../utils/logUtil";
+import ScriptLoader from "../ScriptLoader";
+import { isDefinedAndNotNull } from "../utils/commonUtils";
+
+class Mouseflow {
```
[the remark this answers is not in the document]
sauravlal-rudderstack: @ItsSudip , could you check this?

integrations/Mouseflow/browser.js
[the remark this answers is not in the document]
saikumarrs: If you want to set variables to Mouseflow, then why take the values from the `tags` field?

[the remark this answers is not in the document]
sauravlal-rudderstack: We can pass variables through properties or traits, but by passing through tags we give more flexibility to customers, it's my view.

integrations/Mouseflow/browser.js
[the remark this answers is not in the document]
ItsSudip: As Segment is not implementing this, we are also not supporting this but, We can implement these as configs by taking them from the dashboard. Should we support this? @saikumarrs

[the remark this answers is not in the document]
saikumarrs: I think we should support it, but I'll let @utsabc call the shots.

integrations/Mouseflow/browser.js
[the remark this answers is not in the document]
saikumarrs: Why `tags`?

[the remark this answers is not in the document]
sauravlal-rudderstack: As we are supporting this on some new destinations like Vero, so we think to implement it here.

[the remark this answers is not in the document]
sauravlal-rudderstack: There is nowhere written that we can support tag specifically and in the mouseflow dashboard it simply looks like normal custom variables which we are already adding through traits or properties, so we can remove the support, What's your thoughts should we remove this?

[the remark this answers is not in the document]
saikumarrs: I don't know why we've chosen `tags` as a generic name for sending destination-specific data, but we can send relevant fields.

```
rudderanalytics.track(<event>, <properties>, {
  integrations: {
    Mouseflow: {
      variables: {
        var1: val1,
        var2: val2,
        ...
      }
    }
  }
});
```

@utsabc @ItsSudip you should take a final call on this.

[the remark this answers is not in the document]
sauravlal-rudderstack: Yeah, I'll discuss it tomorrow with Sudip and Utsab.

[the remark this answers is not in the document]
ItsSudip: @sauravlal-rudderstack it is not correct. We should call [this](https://js-api-docs.mouseflow.com/#tagging-a-recording) function for tags. CustomVariables and tags are totally different. I don't know why you were calling `customVariables` in place of `tag`, though while setting the event name as `tag` in line number 85 of track call you were using [this](https://js-api-docs.mouseflow.com/#tagging-a-recording) function.

integrations/Mouseflow/browser.js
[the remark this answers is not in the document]
utsabc: If it does not break our original design we should implement it @sauravlal-rudderstack @ItsSudip

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,105 @@
...
+   * Ref: https://js-api-docs.mouseflow.com/#identifying-a-user
+   * for supporting user traits and tags
+   * Ref: https://js-api-docs.mouseflow.com/#setting-a-custom-variable
+   * @param {Identify} identify
+   */
+  identify(rudderElement) {
+    const { message } = rudderElement;
+    const { traits } = message.context;
+    const email =
+      get(message, "context.traits.email") || get(message, "traits.email");
```
utsabc, line 59: Use optional chaining here

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,105 @@
...
+    const userId = message.userId || email;
```
utsabc, line 60: Why are we passing `email` when `userId` not present. In case of total `anonymous` user where only `anonymousId` is present how will we record?

ItsSudip: It's basically the name of the user. We can send `userId` || `email` || `anonymousId`. WDYT? @utsabc

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,105 @@
...
+    window._mfq.push(["stop"]);
+    if (userId) window._mfq.push(["identify", userId]);
+    window.mouseflow.start();
+    setCustomVariables(traits);
```
utsabc, line 64: Did we check if `traits` is undefined how the function `setcustomVariables` is handling it?

integrations/Mouseflow/utils.js
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable no-underscore-dangle */
+/*
+ * Here, we are iterating each key-value pair of object 'Obj' and
+ * checks if typeof value is string then we pass it as custom variable in mouseflow.
+ */
+const setCustomVariables = (Obj) => {
```
utsabc, line 6: Argument should always be camel case.

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,105 @@
...
+    logger.debug("===In isReady mouseflow===");
+    return !!window._mfq;
+  }
+
+  /*
+   * Add tags
+   * Set custom Variables
+   * Ref: https://js-api-docs.mouseflow.com/#setting-a-custom-variable
+   */
+  addTags(message) {
```
utsabc, line 40: This function can be moved to utils

utsabc (changes requested):

integrations/Mouseflow/browser.js
sauravlal-rudderstack: updated

integrations/Mouseflow/browser.js
sauravlal-rudderstack: updated

integrations/Mouseflow/utils.js
sauravlal-rudderstack: updated

integrations/Mouseflow/browser.js
sauravlal-rudderstack: updated

integrations/Mouseflow/browser.js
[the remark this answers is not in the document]
sauravlal-rudderstack: updated

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,90 @@
...
+  constructor(config) {
+    this.websiteId = config.websiteId;
+    this.name = NAME;
+  }
+
+  init() {
+    logger.debug("===In init mouseflow===");
+    window._mfq = window._mfq || [];
+    ScriptLoader(
+      "mouseflow-integraions",
```
saikumarrs, line 18:
```suggestion
      "mouseflow-integration",
```

integrations/Mouseflow/browser.js
```diff
@@ -0,0 +1,90 @@
...
+
+    const { message } = rudderElement;
+    const { event, properties } = message;
+    if (!event) {
+      logger.error("[mouseflow]: Event name from track call is missing!!===");
+      return;
+    }
+    window._mfq.push(["tag", event]);
+    setCustomVariables(properties);
+    addTags(message);
```
saikumarrs, line 72: Why would anyone want to send one tag as the event name and the rest of them in the integrations object? @ItsSudip @utsabc.

integrations/Mouseflow/browser.js
sauravlal-rudderstack: Updated

integrations/Mouseflow/browser.js
ItsSudip: By default, we are storing the event name as a tag. And apart from that if someone wants to send more tags they can do through the integrations object.

saikumarrs: That doesn't make much sense. Is there a use case for this?

integrations/Mouseflow/browser.js
utsabc: Sounds good

integrations/Mouseflow/browser.js
ItsSudip: There is not any endpoint to send track calls. Segment is sending the properties as key-value pairs. We thought of adding the event name as a tag will help them filter users.

utsabc: You can use tags for filtering/analysis. People might want to associate activity with custom tags, we have designed to allow users flexibility add custom tags.

integrations/Mouseflow/utils.js
```diff
@@ -0,0 +1,35 @@
...
+
+/*
+ * Add tags
+ * Set custom Variables
+ * Ref: https://js-api-docs.mouseflow.com/#setting-a-custom-variable
+ */
+const addTags = (message) => {
+  const tags = get(message, `integrations.${NAME}.tags`);
+  if (isDefinedAndNotNull(tags) && Array.isArray(tags)) {
+    for (let i = 0; i < tags.length; i++) {
```
utsabc, line 28: use `forEach` or `for of` here

integrations/Mouseflow/browser.js
saikumarrs: Treating the event name and tags from the integrations object as tags don't sound like a good idea. Why would anyone want to send tags in two places?

Only either of the following options makes sense to me:
1. Multiple track calls with valid event names with/without any properties to register event names as tags.
2. A single track call without event name but with/without properties and one or more tags in the integrations object.

utsabc: 2. Is not really something we should go towards, event= mouseflow api recommends to set tags based on user activity which correlates more towards our event name

For passing tags in integrations object is allowing users to add custom tag to the session when this activity was  captured as tag.

We can see how we are envisioning this on mouseflow dashboard over a call would be easier to come to some alignment

integrations/Mouseflow/utils.js
sauravlal-rudderstack: updated

integrations/Mouseflow/browser.js
sauravlal-rudderstack: done

ItsSudip (dismissed):

utsabc (dismissed):

integrations/Mouseflow/browser.js
sauravlal-rudderstack: updated the tags to customVariables

ItsSudip (dismissed):

integrations/Mouseflow/browser.js
```diff
@@ -34,33 +34,33 @@ class Mouseflow {
...
   identify(rudderElement) {
+    logger.debug("===In mouseflow Identify===");
     const { message } = rudderElement;
     const { traits } = message.context;
     const email = message.context.traits?.email || message.traits?.email;
-    const userId = message.userId || email;
+    const userId = message.userId || email || message.anonymousId;
     window._mfq.push(["stop"]);
-    if (userId) window._mfq.push(["identify", userId]);
+    if (userId) window.mouseflow.identify(userId);
```
saikumarrs, line 48: Why use the object instead of the queue here?

saikumarrs (dismissed):

integrations/Mouseflow/browser.js
sauravlal-rudderstack: As with queue, we tried but it does not work, and by using object it works, so we take this.

ItsSudip: It is a weird behavior of their SDK. They can't run the function if we use the queue after stopping their SDK. So we have to use the object.

saikumarrs: You are doing it wrong. Identify the user after starting the recording.

saikumarrs: Actually, why are we stopping and starting the recording in the first place?

sauravlal-rudderstack: Sorry my bad, I'll update it

saikumarrs (approved):

PR #595, opened 2022-08-03
feature(event filtering): criteo onetag

mihir-4116:
## Description of the change

- in browser.js file of criteo destination, in track call, logic added in order to map events

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [x] Lint rules pass locally
- [x] The code changed/added as part of this pull request has been covered with tests
- [x] All tests related to the changed code pass in development

### Code review 

- [x]  This pull request has a descriptive title and information useful to a reviewer. There may be a screenshot or screencast attached
- [x] "Ready for review" label attached to the PR and reviewers mentioned in a comment
- [x] Issue from task tracker has a link to this pull request

- - -
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/595)

### DIFF
--- a/integrations/Criteo/browser.js
+++ b/integrations/Criteo/browser.js
@@ -9,8 +9,8 @@ import {
   handlingEventDuo,
   handleListView,
 } from "./utils";
-import { NAME } from "./constants";
-import {getHashFromArrayWithDuplicate} from '../utils/commonUtils';
+import { NAME, supportedEvents } from "./constants";
+import { getHashFromArrayWithDuplicate } from "../utils/commonUtils";
 
 class Criteo {
   constructor(config) {
@@ -121,16 +121,23 @@ class Criteo {
 
     const eventMapping = getHashFromArrayWithDuplicate(this.eventsToStandard);
     const trimmedEvent = event.toLowerCase().trim();
-   
-    if(!eventMapping[trimmedEvent]){
+
+    if (
+      !supportedEvents.includes(trimmedEvent) &&
+      !eventMapping[trimmedEvent]
+    ) {
       logger.debug(`[Criteo] event ${eventType} is not supported`);
       return;
     }
+    let events = [];
+    if (supportedEvents.includes(trimmedEvent)) {
+      events.push(trimmedEvent);
+    } else {
+      events = eventMapping[trimmedEvent];
+    }
 
-    const events = eventMapping[trimmedEvent];
-
-    if(events.length > 0){
-      events.forEach(eventType => {
+    if (events.length > 0) {
+      events.forEach((eventType) => {
         switch (eventType) {
           case "product viewed":
             handleProductView(rudderElement.message, finalPayload);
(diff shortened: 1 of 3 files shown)


integrations/Criteo/constants.js
```diff
@@ -1,8 +1,12 @@
+
 const NAME = "CRITEO";
+
 const CNameMapping = {
   [NAME]: NAME,
   Criteo: NAME,
   criteo: NAME,
 };
 
-export { NAME, CNameMapping };
```
ItsSudip, line 8: This file is not formatted.

integrations/utils/commonUtils.js
```diff
@@ -1,5 +1,4 @@
 import _ from "lodash";
-import logger from "../../utils/logUtil";
```
ItsSudip, line 2: Is this not used anywhere in this file?

mihir-4116: @ItsSudip , no we are not using it anywhere in file

ItsSudip (dismissed):

integrations/Criteo/browser.js
```diff
@@ -116,23 +118,47 @@ class Criteo {
...
+      return;
     }
+    let events = [];
+    if (supportedEvents.includes(trimmedEvent)) {
+      events.push(trimmedEvent);
+    } else {
+      events = eventMapping[trimmedEvent];
+    }
+
+    if (events.length > 0) {
```
saikumarrs, line 139: No need for this check. Array.forEach implicitly handles empty arrays.

integrations/utils/commonUtils.js
```diff
@@ -40,9 +39,12 @@ const getHashFromArrayWithDuplicate = (
         : array[fromKey].trim();
 
       if (hashMap[key]) {
-        hashMap[key].push(array[toKey]);
+        const val = hashMap[key];
+        if (!val.includes(toKey)) {
```
saikumarrs, line 43: What's the purpose of this check?

ItsSudip: We are checking if duplicate elements are added to the array or not. Let's say someone provided same event name mapping twice. Something like:
```
[
          {
            "from": "Downloaded",
            "to": "download"
          },
          {
            "from": "Downloaded",
            "to": "download"
          }
        ]
```
then the Downloaded key will contain the array with ["download", "download"]. To omit this situation we are doing this check.

saikumarrs: In that case, we should not be checking for the presence of `toKey` (which is the string literal `to`).

saikumarrs: Shouldn't it be `if (!val.includes(array[toKey])) {`?

ItsSudip: We decided to do that with help of set but that would be a breaking change.

saikumarrs: I don't understand. It has a logical error.

ItsSudip: Yes. Updated the code.

saikumarrs (approved):

ItsSudip (approved):

PR #657, opened 2022-09-20
feat: enhanced logger module

saikumarrs:
## Description of the change

Added a new logger module and implemented it for Amplitude destination.
A few advantages:
- The logger instance is unique to the module that it is used. We can set `scope` which will be automatically prefixed to the messages. For destinations, it can be the name of the destination.
- Adds a default prefix "RS SDK".
- The message prefixes help the developers to easily identify and filter the messages of the SDK/scope on the browser console.

The plan is to slowly extend this for the rest of the SDK going forward.

Example output:
image

## Type of change

## Related issues

N/A

### Development

- [x] Lint rules pass locally

### Code review 

### DIFF
--- a/utils/logger.js
+++ b/utils/logger.js
@@ -1,0 +1,93 @@
+const LogLevel = {
+  LOG: {
+    value: 0,
+    method: console.log,
+  },
+  INFO: {
+    value: 1,
+    method: console.info,
+  },
+  DEBUG: {
+    value: 2,
+    method: console.debug,
+  },
+  WARN: {
+    value: 3,
+    method: console.warn,
+  },
+  ERROR: {
+    value: 4,
+    method: console.error,
+  },
+};
+
+class Logger {
+  constructor(scope, level) {
+    this.level = +level || LogLevel.ERROR.value;
+    this.scope = scope || "";
+  }
+
+  setLogLevel(levelStr) {
+    this.level = levelStr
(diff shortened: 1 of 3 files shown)


utils/logger.js
```diff
@@ -0,0 +1,107 @@
...
+  },
+  ERROR: {
+    value: 4,
+    method: console.error,
+  },
+};
+
+class Logger {
+  constructor(scope, level) {
+    this.level = +level || LogLevel.ERROR.value;
```
utsabc, line 26: Why are we setting default as error level?

saikumarrs: To avoid noise in the browser console.

utils/logger.js
```diff
@@ -0,0 +1,107 @@
...
+    this.scope = scope || "";
+  }
+
+  setLogLevel(levelStr) {
+    this.level = levelStr
+      ? LogLevel[levelStr.toString().toUpperCase()].value
+      : this.level;
```
MoumitaM, line 33: In case user provides something that is not included in LogLevel array will throw an error while accessing the `.value`. Handle that scenario.

saikumarrs: Good catch. Fixed it.

MoumitaM (approved):

utsabc (approved):

Issue #665, opened 2022-09-26
BUG : All the methods that are called before the `load` method might never get fired


duannx:
**Describe the bug**
When using the SDK via the npm package, the methods that are called before the `load` method might never get fired

**To Reproduce**
Let's say we initiate the analytics like this: 
```ts
const Analytics = {
  instance: null,
  async init() {
    this.instance = await import('rudder-sdk-js')
    // When we did import the SDK, all the methods that are added to `window.rudderanalytics` will be 
    // immediately run but the load method is not called yet. So all the methods will be canceled.
    this.instance.load(WRITE_KEY,DATA_PLANE_URL )
  },
  trackBeforeLoad(methodName: string, ...args: any) {
    if (!window.rudderanalytics) {
      window.rudderanalytics = []
    }
    window.rudderanalytics.push(
      [methodName].concat(Array.prototype.slice.call(args))
    )
  },
  track(...args: Parameters<typeof RudderStackAnalytics.track>) {
    if (!this.instance) return this.trackBeforeLoad('track', ...args)
    return this.instance.track(...args)
  },
}

// Suppose we use the analytics like this
// This method will not be fired
Analytics.track('test')
await Analytics.init()
// This method is fired as expected
Analytics.track('test 2')
```
In the above scenario, the `test` method will not be fired because the analytics is not loaded yet. The reason is that the SDK currently guarantees the `load` call by using `window.rudderanalytics`. If we call the `load` method directly, other methods that are called before it will be ignored.

**Expected behavior**
All the methods that are called before the `load` method should be fired regardless of the way we call the `load` method

**Suggestion**
A workaround is adding the `load` method to `window.rudderanalytics` like this
```
window.rudderanalytics.push(['load', WRITE_KEY,DATA_PLANE_URL])
```
But it is not the natural way when we use an npm package. It only makes sense when we use the SDK from a CDN. 
Because all the methods check `if (!this.loaded) return;` before running, we should put the execution of all the methods after [the `loaded` property is set](https://github.com/rudderlabs/rudder-sdk-js/blob/production/analytics.js#L1009) like this:
```
this.loaded = true
processDataInAnalyticsArray(this)
```
This will guarantee the execution of all methods that are called before the `load` method and provide more future-proof.

**Additional Information:**
- SDK installation: NPM
- NPM Version: 2.15.0

duannx:
Hi @saikumarrs, please take a look at this  🙏🏻

[...]

PR #680, opened 2022-10-12
fix(core): execute buffered requests during load API

MoumitaM:
## Description of the change
- Updated request buffering part during SDK initialization.
- Added a call to execute buffered requests at the end of load API.

## Type of change

## Related issues

> Fix [#665]() 

### Development

- [ ] Lint rules pass locally

### Code review 

- - -
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/680)


analytics.js
```diff
@@ -1265,12 +1276,20 @@ if (isValidArgsArray) {
...
   }
+
+  /**
+   * If defaultMethod is present in the buffered requests execute it.
+   */
+  if (argumentsArray.length > 0 && argumentsArray[0][0] === defaultMethod) {
+    argumentsArray[0].shift();
+    instance[defaultMethod](...argumentsArray[0]);
+    argumentsArray.shift();
+  }
```
saikumarrs, line 1292: Let's not execute the `load` API separately

analytics.js
```diff
@@ -1185,15 +1194,17 @@ class Analytics {
...
+    analytics.toBeProcessedArray.forEach((x) => {
+      const event = [...x];
+      const method = event[0];
+      event.shift();
+      // logger.debug("=====from analytics array, calling method:: ", method)
+      analytics[method](...event);
+    });
+
+    analytics.toBeProcessedArray = [];
+  }
```
saikumarrs, line 1207: Would this work?

```suggestion
  // Process the queue only if the default API is its first entry 
  // or SDK is already loaded
  if (analytics.toBeProcessedArray.length && (analytics.loaded || analytics.toBeProcessedArray[0][0] === defaultMethod) ) {
  
    while (analytics.toBeProcessedArray.length > 0) {
      const event = [...analytics.toBeProcessedArray[0]];
      
      // remove the element from the queue
      analytics.toBeProcessedArray.shift();
      
      const method = event[0];
      event.shift();
      // logger.debug("=====from analytics array, calling method:: ", method)
      analytics[method](...event);
    }
  }
```

analytics.js
```diff
@@ -1034,6 +1035,10 @@ class Analytics {
         } else {
           this.processResponse(200, res);
         }
+        /**
+         * In case load API is called later, execute the buffered requests.
+         */
```
saikumarrs, line 1040:
```suggestion
        // Execute any pending buffered requests
        // (needed if the load call was not previously buffered)
```

analytics.js
```diff
@@ -1048,6 +1053,10 @@ class Analytics {
     } catch (error) {
       handleError(error);
     }
+    /**
+     * In case load API is called later, execute the buffered requests.
+     */
```
saikumarrs, line 1058:
```suggestion
    // Execute any pending buffered requests
    // (needed if the load call was not previously buffered)
```

saikumarrs (changes requested):

analytics.js

tech-goldenphoenix (changes requested):

analytics.js
```diff
@@ -1185,15 +1192,24 @@ class Analytics {
...
-    analytics[method](...event);
-  });
-
-  instance.toBeProcessedArray = [];
+  // Process the queue only if the default API is its first entry
+  // or SDK is already loaded
+  if (
+    analytics.toBeProcessedArray.length &&
+    (analytics.loaded || analytics.toBeProcessedArray[0][0] === defaultMethod)
+  ) {
```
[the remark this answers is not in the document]
MoumitaM: @tech-goldenphoenix , Here `defaultMethod` is `load`. No other event will be processed till `analytics.loaded` is `true`. So, the `analytics.toBeProcessedArray` will still have `event a` and `event b`.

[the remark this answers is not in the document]
duannx: @MoumitaM Let's say the state appears that: 
```js
analytics.loaded = false
analytics.toBeProcessedArray = [['load'], ['track', 'event a'], ['track', 'event b']]
```
The `if` statement will be `true`. So the `while` statement will be executed so these following methods will be fired in synchronous order: 
`['load'] => ['track', 'event a'] => ['track', 'event b']`

But if the `load` method has to load the polyfill, the method `loadAfterPolyfill` will be called asynchronously. And the property `analytics.loaded` will be set to true after all the events fired
`['load'] => ['track', 'event a'] => ['track', 'event b'] => 'load' method load the polyfill asynchronously => analytics.loaded = true`

Based on the above flow, `event a` and `event b` will be ignored because the `analytics.loaded` is not set to true yet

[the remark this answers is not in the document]
MoumitaM: Apologies for the delay. Yes, you are right. Some events might be ignored depending on the polyfill and network strength. Refactored the code. 
Let us know if this works for you.

analytics.js
```diff
@@ -1271,6 +1284,12 @@ if (isValidArgsArray) {
     }
     i += 1;
   }
+
+  if (instance.toBeProcessedArray.length && instance.toBeProcessedArray[0][0] === defaultMethod) {
+    instance.toBeProcessedArray[0].shift();
+    instance[defaultMethod](...instance.toBeProcessedArray[0]);
```
duannx, line 1290: If we execute the `load` methods here. All other methods will be ignored in the normal case when no polyfill is loaded. Because the `instance.toBeProcessedArray` is currently empty 
I think we should double check the test cases too because I see all the tests were passed

duannx (changes requested):

analytics.js
saikumarrs: Yes. @MoumitaM, the polyfill-based load problem should be solved differently.
Let's revert this commit.

duannx: Sorry. It's my bad. Actually, other methods won't be ignored because you make a manual call of `processDataInAnalyticsArray` after calling the `load` method. But it shows another error.
```js
// assume that instance.toBeProcessedArray = [['load', 'abc', 'xyz']]
instance.toBeProcessedArray[0].shift();
// now instance.toBeProcessedArray = [['abc', 'xyz']]
// executing the `load` method
instance[defaultMethod](...instance.toBeProcessedArray[0]);
// the function processDataInAnalyticsArray will be called inside the `load` method. Then it will throw an error since instance.toBeProcessedArray is equal to `[['abc', 'xyz']]`
// this line is executed after the function processDataInAnalyticsArray is called but it's too late
instance.toBeProcessedArray.shift();

```
Another thing is that the code flow is a bit complicated. The function `processDataInAnalyticsArray` is called at 3 different places. It's way hard to follow and maintain. My suggestion is: 
- Remove the `load` method from `instance.toBeProcessedArray` before executing it
- Push all other methods into `instance.toBeProcessedArray` before executing the `load` method
- Only call the `processDataInAnalyticsArray` function inside the `loadAfterPolyfill` function

This will ensure that all the methods will be executed after the `load` method regardless the polyfill is loaded or not.

MoumitaM: Addressed the concerns.

analytics.js
```diff
@@ -1278,7 +1291,11 @@ parseQueryString(window.location.search);
 
 if (isValidArgsArray) argumentsArray.forEach((x) => instance.toBeProcessedArray.push(x));
 
-processDataInAnalyticsArray(instance);
+if (instance.toBeProcessedArray.length && instance.toBeProcessedArray[0][0] === defaultMethod) {
```
saikumarrs, line 1294: Add documentation here.

saikumarrs, line 1294: Extract the load event in the while block above and execute here directly.

analytics.js
```diff
@@ -1185,15 +1192,21 @@ class Analytics {
...
-    const event = [...x];
-    const method = event[0];
-    event.shift();
-    // logger.debug("=====from analytics array, calling method:: ", method)
-    analytics[method](...event);
-  });
-
-  instance.toBeProcessedArray = [];
+  // Process the queue only if the default API is its first entry
+  // or SDK is already loaded
```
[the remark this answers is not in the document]

saikumarrs (approved):

duannx (approved):
LGTM!

PR #685, opened 2022-10-17
feat: add utility to generate integration common names automatically

saikumarrs:
## Description of the change

Added a utility to auto-generate the generic common names for an integration given its display name. For integrations where this default common names map is not enough, we can still define additional entries in the `CNameMapping` object in `constants.js`.

The default common name mapping is built by converting the display name to lowercase. Hence, when comparing any name from the integrations object, it is also first converted to lowercase.

Additionally, refactored the `IsEventBlackListed` function to remove redundant common names resolution logic.

This feature has two benefits:
- Reduced SDK size
- Less error prone w.r.t common names map definition.

It is currently implemented as an example for the "Adobe Analytics" destination.

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

- - -
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/685)


utils/utils.js
```diff
@@ -286,12 +286,17 @@ function getRevenue(properties, eventName) {
 function tranformToRudderNames(integrationObject) {
   Object.keys(integrationObject).forEach((key) => {
     if (integrationObject.hasOwnProperty(key)) {
-      if (commonNames[key]) {
-        integrationObject[commonNames[key]] = integrationObject[key];
+      const sanitizedIntgName = key.toLowerCase().trim();
```
bardisg, line 289: Not sure about this one, as I do not know the permutations of the key values.

It seems like only the display name is cast to lowercase in the utility method, for example NAME or any hardcoded CNameMapping coould have uppercase values.

If the key value in this context is alwys the display name this is fine elsewise toLowerCase would not work for all cases

utils/utils.js
```diff
@@ -286,12 +286,17 @@ function getRevenue(properties, eventName) {
...
+      if (commonNames[sanitizedIntgName]) {
+        integrationObject[commonNames[sanitizedIntgName]] =
+          integrationObject[key];
       }
       if (key != "All") {
         // delete user supplied keys except All and if except those where oldkeys are not present or oldkeys are same as transformed keys
-        if (commonNames[key] != undefined && commonNames[key] != key) {
+        if (
+          commonNames[sanitizedIntgName] != undefined &&
+          commonNames[sanitizedIntgName] != key
```
bardisg, line 298: we should be using !==

utils/utils.js
saikumarrs: Yes, the keys in the common names map are always a combination of the display name. 

The casing is made to lower on both sides to eliminate any play w.r.t that. One would ideally override (rather, add new ones) the cname mapping for other combinations (ex: "Customer.io" could have other hardcoded names defined such as "customer-io", "kustomerio" (with a "K"), etc).

However, I've now fixed the bug (converted the case of even the hard-coded cname mappings)

bardisg: got it. LGTM

utils/utils.js
bardisg: I see

MoumitaM (dismissed):

shrouti1507 (approved):

integrations/utils/commonUtils.js
```diff
@@ -163,6 +163,40 @@ function getDestinationExternalID(message, type) {
...
+    cNamesMap[words.join("")] = nameSanitized;
+  }
+
+  // add the hard-coded common names
+  if (CNameMapping) {
+    // convert the keys to lower case to
+    // match the existing common names
+    Object.keys(CNameMapping).forEach((name) => {
+      const sanitizedName = name.toLowerCase().trim();
+      cNamesMap[sanitizedName] = CNameMapping[name];
```
MoumitaM, line 193: `cNamesMap[sanitizedName] = CNameMapping[name];` this will create duplicate records as we are not removing the existing key-value. Is this expected?

MoumitaM (commented):
With this change developers now have to keep in mind what combination will be generated by the SDK and also they have to add common names that don't fall into this category. Previously they used to write the all possible combination by themselves. Just thinking out loud, wouldn't this create more confusion and lead to more mistakes?

integrations/utils/commonUtils.js
saikumarrs: It'll override if a key is already present.

saikumarrs:
> With this change developers now have to keep in mind what combination will be generated by the SDK and also they have to add common names that don't fall into this category. Previously they used to write the all possible combination by themselves. Just thinking out loud, wouldn't this create more confusion and lead to more mistakes?

This utility should tackle more than 90% (100% in some cases) of the common names for all the destinations. I believe we can add documentation to clarify the combinations that are taken care of automatically. Even if someone makes a mistake, they'd just redundantly override a common name.

MoumitaM (approved):

MoumitaM (approved):

MoumitaM:
@saikumarrs , We can not merge this feature. 
`tranformToRudderNames` is a common function to transform user provided name. But we have modified it. So the existing implementation might break. 
Ex:
```
{
   integrations: {
       All: false,
       "Amplitude": true
   }
}
```
Now there is no mapping b/w "amplitude" to "AM" in common names and modified integrations object will become:
```
{
   integrations: {
       All: false,
       "Amplitude": true
   }
}
```
So no event will reach `Amplitude`.

saikumarrs:
> @saikumarrs , We can not merge this feature. `tranformToRudderNames` is a common function to transform user provided name. But we have modified it. So the existing implementation might break. Ex:
> 
> ```
> {
>    integrations: {
>        All: false,
>        "Amplitude": true
>    }
> }
> ```
> 
> Now there is no mapping b/w "amplitude" to "AM" in common names and modified integrations object will become:
> 
> ```
> {
>    integrations: {
>        All: false,
>        "Amplitude": true
>    }
> }
> ```
> 
> So no event will reach `Amplitude`.

Really good catch!
I've fixed this issue. Please review.

MoumitaM (changes requested):
We need to update the `IsEventBlackListed` function.
```
IsEventBlackListed(eventName, intgName) {
    if (!eventName || !(typeof eventName === "string")) {
      return false;
    }
    const sdkIntgName = commonNames[intgName];
    const intg = this.clientIntegrations.find(
      (intg) => intg.name === sdkIntgName
    );
```
This `intgName` is the name of an integration mentioned inside the constructor fn. 
now there is no mapping b/w ADOBE_ANALYTICS -> ADOBE_ANALYTICS. So this feature will not work for Adobe.
We need to update the logic with:
```
const sdkIntgName = commonNames[intgName] || commonNames[intgName.toLowerCase().trim()];
```

saikumarrs:
> 

It's good that you brought it up. We don't need the common name resolution at all, as the names in the destination instance object and the integrations object are already in the rudder names format.
Fixed it.

saikumarrs:
> We need to update the `IsEventBlackListed` function.
> 
> ```
> IsEventBlackListed(eventName, intgName) {
>     if (!eventName || !(typeof eventName === "string")) {
>       return false;
>     }
>     const sdkIntgName = commonNames[intgName];
>     const intg = this.clientIntegrations.find(
>       (intg) => intg.name === sdkIntgName
>     );
> ```
> 
> This `intgName` is the name of an integration mentioned inside the constructor fn. now there is no mapping b/w ADOBE_ANALYTICS -> ADOBE_ANALYTICS. So this feature will not work for Adobe. We need to update the logic with:
> 
> ```
> const sdkIntgName = commonNames[intgName] || commonNames[intgName.toLowerCase().trim()];
> ```


It's good that you brought it up. We don't need the common name resolution at all, as the names in the destination instance object and the integrations object are already in the rudder names format.
Fixed it.

MoumitaM (approved):

integrations/integration_cname.js
```diff
@@ -135,6 +136,7 @@ const commonNames = {
   ...Refiner,
   ...YandexMetrica,
   ...Qualaroo,
+  all: "All",
```
MoumitaM, line 139:
```suggestion
  All: "All",
```

MoumitaM (changes requested):
Found another bug, `All: true/false` was getting converted to `All: undefined` and all the destinations started loading.

integrations/integration_cname.js
saikumarrs: The fix is applied in a different location.

MoumitaM (approved):

shrouti1507 (approved):

PR #723, opened 2022-11-07
feature: posthog option to not load core sdk

mihir-4116:
## Description of the change

- in this pr we are giving option to load sdk or not for POSTHOG destination
- if user already loaded sdk then they just need to send below code snipped so that we will not load sdk from our end

` rudderanalytics.load(
        "write key",
        "DATA_PLANE_URL",
        {
          configUrl: "CONFIG_URL",
          logLevel: "DEBUG",
          integrations: {
            All: true,
            POSTHOG: {
             useGloballyAvailableIntegration: false,
            },
          },
       });`

- if `loadIntegration: false` isn't there then we will load sdk from our end

## Type of change
- [ ✅] New feature (non-breaking change that adds functionality)

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

- [ ✅] "Ready for review" label attached to the PR and reviewers mentioned in a comment

- - -
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/723)


bardisg:
How do we ensure that:
- the user has the correct version loaded
- the integration is already available in expected window property window.posthog before we use it
- is this something taht we will support for all integrations? Has the product team agreed on this? Please confirm with Arnub
- have we tested the scenario where useGlobalyAvailableIntegration is set to true but it was never available? What happens in this case

mihir-4116:
> How do we ensure that:
> 
> * the user has the correct version loaded
> * the integration is already available in expected window property window.posthog before we use it
> * is this something taht we will support for all integrations? Has the product team agreed on this? Please confirm with Arnub
> * have we tested the scenario where useGlobalyAvailableIntegration is set to true but it was never available? What happens in this case

cc : @ItsSudip

ItsSudip:
> How do we ensure that:
> 
> * the user has the correct version loaded
> * the integration is already available in expected window property window.posthog before we use it
> * is this something that we will support for all integrations? Has the product team agreed on this? Please confirm with @arnab-p
> * have we tested the scenario where useGlobalyAvailableIntegration is set to true but it was never available? What happens in this case

@bardisg we have a similar implementation in [VWO destination](https://github.com/rudderlabs/rudder-sdk-js/blob/4ef1348ddcb06fee19dc9ec6228f6040c5244e32/integrations/VWO/browser.js#L27). 
For the version we can update the documentation i.e. if the customer is using this feature then they need to make sure they are using the same version in which our implementation is on.
We are using isloaded function to check if the sdk is loaded correctly or not.
I didn't get the last concern. Do we have any variable in our js SDK with this same name?

mihir-4116:
i tested for different cases : 

case 1 : 
if user load posthog sdk and sends `POSTHOG: { loadIntegration: false }` then we will not load sdk from our end but sdk is loaded successfully, also user is able to send calls through rudder-sdk

case 2 : 
if user dosen't load posthog sdk and sends `POSTHOG: { loadIntegration: false }` then that particular integration will not loaded (which is expected)

case 3 :
if user load posthog sdk and dosen't send `POSTHOG: { loadIntegration: false }` then in console below error will popped up, but sdk is loaded and user will be able to send data
Screenshot 2022-11-15 at 12 12 30 PM

- i didn't found any older version of posthog sdk, whenever we create an account in posthog, posthog will give us latest version of sdk and mostly it will not change

cc : @bardisg @ItsSudip

bardisg (approved):

integrations/Posthog/browser.js
```diff
@@ -44,76 +44,79 @@ class Posthog {
...
-            n++
-          )
-            g(u, o[n]);
-          e._i.push([i, s, a]);
-        }),
-        (e.__SV = 1));
-    })(document, window.posthog || []);
-
     const { POSTHOG } = this.analytics.loadOnlyIntegrations;
+    if (POSTHOG && !POSTHOG.loadIntegration) {
```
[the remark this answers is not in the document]
mihir-4116: updated

ItsSudip (approved):

integrations/Posthog/browser.js
```diff
@@ -44,76 +44,81 @@ class Posthog {
...
-          e._i.push([i, s, a]);
-        }),
-        (e.__SV = 1));
-    })(document, window.posthog || []);
-
     const { POSTHOG } = this.analytics.loadOnlyIntegrations;
+    if (POSTHOG && !POSTHOG.useGloballyAvailableIntegration) {
+      logger.debug(
+        "===[POSTHOG]: useGloballyAvailableIntegration flag is disabled==="
+      );
```
MoumitaM, line 51: The name `useGloballyAvailableIntegration` is suggesting the opposite thing meaning when `useGloballyAvailableIntegration` is true then you should not load the integration because it is globally available.

MoumitaM (changes requested):

integrations/Posthog/browser.js
MoumitaM, line 51:
```suggestion
    if (POSTHOG && POSTHOG.useGloballyAvailableIntegration === true) {
      logger.debug(
        "===[POSTHOG]: useGloballyAvailableIntegration flag is enabled==="
      );
```

mihir-4116: updated

MoumitaM (approved):

ItsSudip (approved):

lewisflude:
@mihir-4116 hey! I appreciate this PR is really old at this point, but was curious if this work exists in any of the official documentation? This seemed to fix a really gnarly bug I was facing

PR #725, opened 2022-11-08
Revert "feat: add utility to generate integration common names automatically"

saikumarrs:
Reverts rudderlabs/rudder-sdk-js#685

Due to the import structure differences b/w v1 and v1.1 SDK architectures, if the changes are directly pulled into v1.1 SDK, it incurs a heavy SDK size cost. Hence, for the time being, reverting the changes to redo the approach.

- - -
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/725)

PR #732, opened 2022-11-16
feat(destination): onboard satismeter

anantjain45823:
## Description of the change

> Onboarding a new destination SatisMeter with Identify and Track Call. Have made a call with writeKey only inside init(), to help to record the survey responses on the destination even without identify call i.e. as anonymous user. 

## Type of change

## Related issues

> Fix [#1]() 

### Development

- [ ] Lint rules pass locally

### Code review 

- [x] "Ready for review" label attached to the PR and reviewers mentioned in a comment

- - -
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/732)

### DIFF
--- a/integrations/Satismeter/browser.js
+++ b/integrations/Satismeter/browser.js
@@ -1,36 +1,43 @@
 /* eslint-disable no-var */
 /* eslint-disable no-underscore-dangle */
 /* eslint-disable class-methods-use-this */
-import logger from "../../utils/logUtil";
 import { NAME } from "./constants";
+import Logger from "../../utils/logger";
+import { recordSatismeterEvents } from "./util";
 import { LOAD_ORIGIN } from "../ScriptLoader";
 
+const logger = new Logger(NAME);
 class Satismeter {
   constructor(config, analytics) {
-    logger.debug("in satismetwr const");
+    this.analytics = analytics;
     if (analytics.logLevel) {
       logger.setLogLevel(analytics.logLevel);
     }
     this.name = NAME;
     this.writeKey = config.writeKey;
+    this.recordSatismeterEvents = config.recordSatismeterEvents;
+    this.eventsToStandard = config.eventsToStandard;
+    this.updateEventNames = config.updateEventNames;
+    this.eventsList = config.eventsList;
   }
 
   init() {
     logger.debug("===In init Satismeter===");
     (function () {
-      var script = document.createElement("script");
-      var parent = document.getElementsByTagName("script")[0].parentNode;
       window.satismeter =
         window.satismeter ||
         function () {
           (window.satismeter.q = window.satismeter.q || []).push(arguments);
         };
       window.satismeter.l = 1 * new Date();
+      var script = document.createElement("script");
+      var parent = document.getElementsByTagName("script")[0].parentNode;
       script.async = 1;
       script.src = "https://app.satismeter.com/js";
-      script.setAttribute("data-loader", LOAD_ORIGIN);
-      parent.appendChild(script);
+      script.setAttribute("data-loader", LOAD_ORIGIN),
+        parent.appendChild(script);
     })();
+
     window.satismeter({
       writeKey: this.writeKey,
     });
@@ -43,6 +50,13 @@ class Satismeter {
 
   isReady() {
     logger.debug("===In isReady Satismeter===");
+    if (this.recordSatismeterEvents) {
+      recordSatismeterEvents(
+        this.updateEventNames,
+        this.eventsList,
+        this.eventsToStandard
+      );
+    }
     return !!window.satismeter;
   }
 
@@ -73,9 +87,14 @@ class Satismeter {
   track(rudderElement) {
     logger.debug("===In Satismeter track===");
     const { message } = rudderElement;
-    const { event } = message;
+    const { event, context } = message;
     if (!event) {
       logger.error("[Satismeter]:: event is required for track call");
+    }
+    const integrationName = context.integration?.name;
+    if (integrationName && integrationName.toLowerCase() === "satismeter") {
+      logger.debug(`[Satismeter]:: dropping callback event: ${event}`);
+      return;
     }
     window.satismeter("track", { event });
   }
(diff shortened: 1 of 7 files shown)


MoumitaM (dismissed):

integrations/Satismeter/browser.js
```diff
@@ -0,0 +1,84 @@
+/* eslint-disable no-var */
+/* eslint-disable no-underscore-dangle */
+/* eslint-disable class-methods-use-this */
+import logger from "../../utils/logUtil";
```
ItsSudip, line 4: We should use `Logger` from `../../utils/logger`

integrations/Satismeter/browser.js
```diff
@@ -0,0 +1,84 @@
...
+import { NAME } from "./constants";
+import { LOAD_ORIGIN } from "../ScriptLoader";
+
+class Satismeter {
+  constructor(config, analytics) {
+    logger.debug("in satismetwr const");
+    if (analytics.logLevel) {
+      logger.setLogLevel(analytics.logLevel);
+    }
+    this.name = NAME;
```
ItsSudip, line 14: We should use the new common name module. For reference: https://github.com/rudderlabs/rudder-sdk-js/pull/685, https://github.com/rudderlabs/rudder-sdk-js/pull/725

integrations/Satismeter/browser.js
```diff
@@ -0,0 +1,84 @@
...
+        window.satismeter ||
+        function () {
+          (window.satismeter.q = window.satismeter.q || []).push(arguments);
+        };
+      window.satismeter.l = 1 * new Date();
+      script.async = 1;
+      script.src = "https://app.satismeter.com/js";
+      script.setAttribute("data-loader", LOAD_ORIGIN);
+      parent.appendChild(script);
+    })();
```
ItsSudip, line 33: We should use the `scriptLoader` function here.

integrations/Satismeter/browser.js
```diff
@@ -0,0 +1,84 @@
...
+  isReady() {
+    logger.debug("===In isReady Satismeter===");
+    return !!window.satismeter;
+  }
+
+  identify(rudderElement) {
+    logger.debug("===In Satismeter identify===");
+    const { message } = rudderElement;
+    const { traits } = message.context;
+    const userId = message.userId || traits?.userId;
```
ItsSudip, line 53: Shouldn't we also track the anonymous users also? Let's provide an option on the dashboard to track or not to track anonymous users and depending on that we can implement this.

integrations/Satismeter/browser.js
```diff
@@ -0,0 +1,84 @@
...
+    if (!userId) {
+      logger.error("[Satismeter]:: userId is not found for Identify Call");
+    }
+    if (!traits?.createdAt) {
+      window.satismeter({
+        writeKey: this.writeKey,
+        userId,
+        type: "identify",
+        traits: { ...traits, createdAt: message.sentAt },
```
ItsSudip, line 62: We should check if `message.sentAt` is available or not.

integrations/Satismeter/browser.js
```diff
@@ -0,0 +1,84 @@
...
+      });
+    }
+    window.satismeter({
+      writeKey: this.writeKey,
+      userId,
+      type: "identify",
+      traits,
+    });
```
ItsSudip, line 70: It should go to the else part of the above if condition.

ItsSudip (changes requested):

integrations/Satismeter/browser.js
anantjain45823: We are already tracking anonymous users in such a way that if track call is made without identify call then it will be anonymous user otherwise details could be taken from identify call

ItsSudip: Didn't get it. We are not providing any user traits through the track call.

integrations/Satismeter/browser.js
anantjain45823: Unable to understand this as the changes already were reverted at the destination m
odule level

anantjain45823: Unable to understand this as changes were in #725 at destination module level.

integrations/Satismeter/browser.js
anantjain45823: Actually while the loading the script we are sending a request with the writeKey which will be helping it to track a response even an anonymous one. But if we are making an identify call first and then track call then it captures the details from the identify call made previously through cookies and all.

integrations/Satismeter/browser.js
anantjain45823: I don’t think we can as scriptLoader is mainly for those which have an link for their script whereas in this destination's case we are not getting any link its a raw script which we have to run

anantjain45823: The script is too general to be used and on trying it as well it didn't work. They are using some internal Functioning due which ScriptLoader can not work.

integrations/Satismeter/browser.js
ItsSudip: Yes, I didn't know it was reverted. So no need to implement this.

integrations/Satismeter/browser.js
```diff
@@ -0,0 +1,103 @@
...
+  track(rudderElement) {
+    logger.debug("===In Satismeter track===");
+    const { message } = rudderElement;
+    const { event, context } = message;
+    if (!event) {
+      logger.error("[Satismeter]:: event is required for track call");
+    }
+    const integrationName = context.integration?.name;
+    if (integrationName && integrationName.toLowerCase() === "satismeter") {
+      logger.debug(`[Satismeter]:: dropping callback event: ${event}`);
```
yashasvibajpai, line 96: We could have used the info log here, as it seems to be more of an information related to the triggered callback.

(Non blocking)

anantjain45823: Yeah makes much more sense.

ItsSudip (approved):

PR #743, opened 2022-11-25
feat/axeptio: onboard web device mode destination axeptio

ujjwal-ab:
## PR Description

Onboard new web device mode destination Axeptio

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security

- - -
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/743)

### DIFF
--- a/integrations/Axeptio/browser.js
+++ b/integrations/Axeptio/browser.js
@@ -1,14 +1,21 @@
 /* eslint-disable no-underscore-dangle */
 /* eslint-disable class-methods-use-this */
 import { NAME } from "./constants";
-import logger from "../../utils/logUtil";
+import Logger from "../../utils/logger";
 import { LOAD_ORIGIN } from "../ScriptLoader";
 import makeACall from "./utils";
 
+const logger = new Logger(NAME);
+
 class Axeptio {
-  constructor(config) {
-    this.clientId = config.clientId;
+  constructor(config, analytics) {
+    this.analytics = analytics;
+    if (analytics.logLevel) {
+      logger.setLogLevel(analytics.logLevel);
+    }
     this.name = NAME;
+    this.clientId = config.clientId;
+    this.toggleToActivateCallback = config.toggleToActivateCallback;
   }
 
   loadScript() {
@@ -37,7 +44,9 @@ class Axeptio {
 
   isReady() {
     logger.debug("===In isReady Axeptio===");
-    this.recordAxeptioEvents();
+    if (this.toggleToActivateCallback) {
+      this.recordAxeptioEvents();
+    }
     return !!window.__axeptioSDK;
   }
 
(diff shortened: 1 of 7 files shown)


integrations/Axeptio/browser.js
```diff
@@ -0,0 +1,60 @@
+/* eslint-disable no-underscore-dangle */
+/* eslint-disable class-methods-use-this */
+import { NAME } from "./constants";
+import logger from "../../utils/logUtil";
```
ItsSudip, line 4: We should update the new `logger` here. You can take reference from amplitude.

integrations/Axeptio/browser.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+
+  isLoaded() {
+    logger.debug("===In isLoaded Axeptio===");
+    return !!window.__axeptioSDK && typeof window.__axeptioSDK === "object";
+  }
+
+  isReady() {
+    logger.debug("===In isReady Axeptio===");
+    this.recordAxeptioEvents();
```
ItsSudip, line 40: We should provide an option in the dashboard to check if the customer wants to use the callback functionality.

integrations/Axeptio/browser.js
```diff
@@ -0,0 +1,60 @@
...
+    return !!window.__axeptioSDK;
+  }
+
+  // this fucntion is used to record the triggered axeptio events through callback
+  recordAxeptioEvents() {
```
ItsSudip, line 45: Are we not filtering out the events? I mean we are setting callbacks for all the events by default?

ItsSudip (changes requested):

integrations/Axeptio/browser.js
ujjwal-ab: Yeah, we can do that. But, as the callback is the only supported feature here. Will it make sense?

integrations/Axeptio/browser.js
ujjwal-ab: Yeah, as we are not totally sure about the event's name. We haven't added any filters and the callback is set to all the triggered events. One check is there for the payload. If the payload is undefined we are only sending the event name else both.

integrations/Axeptio/browser.js
ItsSudip: Yes. Someone may only want to load the axeptio SDK.

integrations/Axeptio/browser.js
ItsSudip: Okay let's support all the events later we can add some filters as an enhancement.

integrations/Axeptio/browser.js
ujjwal-ab: Updated

integrations/Axeptio/browser.js
ujjwal-ab: Updated

ItsSudip (approved):

sandeepdsvs (approved):

aashishmalik (approved):
LGTM

MoumitaM (approved):

PR #747, opened 2022-12-01
feat: onboard ga4 hybrid mode

mihir-4116:
## PR Description

- in `analytics.js` file, when integration is ready we are preparing the data for integrations obj for particular destination
- to get the data from destination we can use a function called `getDataForIntegrationsObject` 
- once the data is prepared we will merge the customer integrationsObj with our integrationsData obj
- we are merging in a way that if there is a conflict in keys between both object then customer integrations object key should override integrationsData object key (customer key will be given first priority)

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security

- - -
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/747)


analytics.js
```diff
@@ -307,6 +327,7 @@ class Analytics {
         const destConfig = intg.config;
         intgInstance = new intgClass(destConfig, self);
         intgInstance.init();
+        this.prepareDataForIntegrationsObj(intg, intgInstance);
```
saikumarrs, line 330: This may not be the right place to get the data from the integration. We have to wait for it to be loaded and ready.

integrations/GA4/utils.js
```diff
@@ -241,11 +240,9 @@ const proceedCloudMode = (rudderElement, measurementId) => {
   let sessionId;
   if (cookieObj[`_ga_${measurementIdArr[1]}`]) {
     sessionId = cookieObj[`_ga_${measurementIdArr[1]}`].split(".");
-    const GA4 = { sessionId: sessionId[2] };
-    payload.message.integrations = { All: true, GA4 };
-    return payload;
+    return { sessionId: sessionId[2] };
```
saikumarrs, line 243: This method should return the session ID value, not the wrapping object. The packaging should be done at the level above.

integrations/GA4/browser.js
```diff
@@ -258,4 +258,8 @@ export default class GA4 {
       });
     });
   }
+
+  getDataForIntegrationsObject() {
+    return getGa4SessionId(this.measurementId);
```
saikumarrs, line 263: Let's return in this format,
```
return {
  <DISPLAY_NAME>: {
    <key value pairs>
  }
}
```

analytics.js
```diff
@@ -833,11 +854,16 @@ class Analytics {
...
+        All: true,
+      };
 
       // structure user supplied integrations object to rudder format
       tranformToRudderNames(clientSuppliedIntegrations);
-      rudderElement.message.integrations = clientSuppliedIntegrations;
+      rudderElement.message.integrations = {
+        ...this.integrationsData,
+        ...clientSuppliedIntegrations,
+      };
```
saikumarrs, line 866: This should be done right before sending it to the cloud mode. Move this after Ln 921. Also, naming transformation should happen for this to work (this is happening in Ln 921).

saikumarrs, line 866: Simple object merge will not work for all the cases. You should use the lodash merge utility.

ex:
```
integrationsData = {
  "Google Analytics 4": {
    "sessionId": "123"
  }
}
------------------------------

rudderanalytics.track("Some event", {}, {
  integrations: {
    All: true
    "Google Analytics 4": {
      "some other data": "abc"
    }
  }
});
------------------------------
This will result in:
{
  type: "track",
  event: "Some event",
  ...
  integrations: {
    All: true
    "Google Analytics 4": {
      "some other data": "abc"
    }
  }
}
but what we want is:
{
  type: "track",
  event: "Some event",
  ...
  integrations: {
    All: true
    "Google Analytics 4": {
      "sessionId": "123",
      "some other data": "abc"
    }
  }
}

```
CC @MoumitaM

analytics.js
```diff
@@ -276,6 +282,20 @@ class Analytics {
...
+   *
+   * @param {*} integration
+   * @param {*} integrationInstance
+   * @memberof Analytics
+   */
+  prepareDataForIntegrationsObj(integration, integrationInstance) {
+    if (integration.name === "GA4" && integrationInstance.isReady()) {
+      this.integrationsData[integration.name] =
+        integrationInstance.getDataForIntegrationsObject();
+    }
```
saikumarrs, line 296: There is no need to be explicit for GA4. You should do this for all integrations.

Rough logic:
```
integrationInstances.forEach((intgInstance) => {
  if (intgInstance["getDataForIntegrationsObject"]) {
    // As each integration will have a unique display name, there is no need for deep object merge
    this.integrationsData = { ...this.integrationsData, ...intgInstance["getDataForIntegrationsObject"]() };
  }
});
```

saikumarrs (changes requested):

analytics.js
mihir-4116: updated

analytics.js
mihir-4116: updated

mihir-4116: updayed

integrations/GA4/utils.js
mihir-4116: updated

integrations/GA4/browser.js
mihir-4116: updated

integrations/GA4/browser.js
```diff
@@ -260,6 +260,10 @@ export default class GA4 {
   }
 
   getDataForIntegrationsObject() {
-    return getGa4SessionId(this.measurementId);
+    return {
+      GA4: {
```
saikumarrs, line 264: Is `GA4` the display name? We should use display name.

mihir-4116: updated

saikumarrs (approved):

ItsSudip (dismissed):
LGTM. One doubt, in some situations we are sending empty strings in place of sessionId is this intentional?

mihir-4116:
> 

it's the case where we don't find sessionId from cookie, from cloud mode it will be ignored since it's empty

integrations/GA4/utils.js
```diff
@@ -236,6 +236,13 @@ const getGa4SessionId = (measurementId) => {
     sessionId = Cookie.get(`_ga_${measurementIdArr[1]}`).split(".");
     return sessionId[2];
   }
+  if (
+    window.localStorage &&
+    window.localStorage.getItem(`_ga_${measurementIdArr[1]}`)
+  ) {
+    sessionId = window.localStorage.getItem(`_ga_${measurementIdArr[1]}`);
+    return sessionId[2];
+  }
```
saikumarrs, line 245: Use the storage module instead.

saikumarrs (changes requested):

integrations/GA4/utils.js
mihir-4116: @saikumarrs As we discussed before, Storage.getItem() function will throw an error because we are parsing response. is there any other function that you are referring to?

`SyntaxError: Unexpected token 'G', "GS1.1.1669"... is not valid JSON`

analytics.js
```diff
@@ -918,8 +921,22 @@ class Analytics {
 
       // convert integrations object to server identified names, kind of hack now!
       transformToServerNames(rudderElement.message.integrations);
+
+      // Removing true values from clientSuppliedIntegrations except All
+      clientSuppliedIntegrations = Object.keys(clientSuppliedIntegrations)
```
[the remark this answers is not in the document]
mihir-4116: updated

analytics.js
```diff
@@ -923,7 +923,9 @@ class Analytics {
       transformToServerNames(rudderElement.message.integrations);
 
       // Removing true values from clientSuppliedIntegrations except All
-      clientSuppliedIntegrations = Object.keys(clientSuppliedIntegrations)
+      const tempClientSuppliedIntegrations = Object.keys(
+        clientSuppliedIntegrations
+      )
         .filter(
           (integration) =>
             integration !== "All" &&
```
saikumarrs, line 931: We shouldn't be filtering the `All` entry.

mihir-4116: updated

integrations/GA4/utils.js
mihir-4116: updated

analytics.js
```diff
@@ -926,11 +926,12 @@ class Analytics {
...
-        .filter(
-          (integration) =>
-            integration !== "All" &&
-            clientSuppliedIntegrations[integration] !== true
-        )
+        .filter((integration) => {
+          if (integration !== "All") {
+            return clientSuppliedIntegrations[integration] !== true;
+          }
+          return integration;
```
saikumarrs, line 933: This should be `true`, right?

integrations/GA4/utils.js
```diff
@@ -236,11 +237,8 @@ const getGa4SessionId = (measurementId) => {
     sessionId = Cookie.get(`_ga_${measurementIdArr[1]}`).split(".");
     return sessionId[2];
   }
-  if (
-    window.localStorage &&
-    window.localStorage.getItem(`_ga_${measurementIdArr[1]}`)
-  ) {
-    sessionId = window.localStorage.getItem(`_ga_${measurementIdArr[1]}`);
+  if (Store.get(`_ga_${measurementIdArr[1]}`)) {
+    sessionId = Store.get(`_ga_${measurementIdArr[1]}`);
```
saikumarrs, line 241: Both from Cookie and Local Storage modules read only once!

saikumarrs (dismissed):

analytics.js
mihir-4116: Since All key value will always set to true, it will always return true

saikumarrs: I mean, `integration` is the name of the integration, which is a string. We should be returning `true` instead to avoid filtering it. It should still work as all non-empty strings evaluate to truthy values but let's still fix it.

mihir-4116: updated

integrations/GA4/utils.js
mihir-4116: updated

saikumarrs (dismissed):

ItsSudip (dismissed):

analytics.js
```diff
@@ -894,6 +922,27 @@ class Analytics {
       // convert integrations object to server identified names, kind of hack now!
       transformToServerNames(rudderElement.message.integrations);
 
+      // Removing true values from clientSuppliedIntegrations except All
+      const tempClientSuppliedIntegrations = Object.keys(
+        clientSuppliedIntegrations
+      )
+        .filter((integration) => {
+          if (integration !== "All") {
+            return clientSuppliedIntegrations[integration] !== true;
```
MoumitaM, line 931: This logic will also filter other integrations from the client-supplied integration that are set to true. We only want that for tempIntegrationsData i.e. GA4. Please modify the logic.

analytics.js
```diff
@@ -894,6 +922,27 @@ class Analytics {
       // convert integrations object to server identified names, kind of hack now!
       transformToServerNames(rudderElement.message.integrations);
 
+      // Removing true values from clientSuppliedIntegrations except All
```
MoumitaM, line 925: Please add elaborated comments to explain the logic.

MoumitaM (changes requested):

analytics.js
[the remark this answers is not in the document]
mihir-4116: it's already updated and generic approach is applied there

analytics.js
mihir-4116: no, this logic is required because if some integrations are set to true and for that if data is required for that integration inside integrationsObj then SDK global object should take priority. we should override default true value and set the data if required this is already discussed with @saikumarrs . this is an edge case that we are handling here?

you can take reference of test cases here :
https://www.notion.so/rudderstacks/Test-Results-3f723a53a98f4f3183a48bfa086afa65

saikumarrs: @mihir-4116 Correct, but we should do it only for those integrations that have data defined in the global integrations data object. Otherwise, we'll be messing up with the event filtering logic.

analytics.js
mihir-4116: updated

analytics.js
mihir-4116: updated

analytics.js
```diff
@@ -922,22 +922,25 @@ class Analytics {
       // convert integrations object to server identified names, kind of hack now!
       transformToServerNames(rudderElement.message.integrations);
 
-      // Removing true values from clientSuppliedIntegrations except All
+      /**
+       * Cloning the global analytics integrationsData object
+       * Filtering the integrations which are not a part of integrationsData object or value set to false
+       * merging tempClientSuppliedIntegrations and tempIntegrationsData using lodash merge method
```
saikumarrs, line 928: This line is redundant.

saikumarrs (dismissed):

analytics.js
mihir-4116: redundant comments removed

saikumarrs (approved):

MoumitaM (approved):

ItsSudip (approved):

PR #817, opened 2023-01-01
feat(braze): make logging configurable

aashishmalik:
## PR Description

Make logging configurable in braze.

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security

- - -
This change is [Reviewable](https://reviewable.io/reviews/rudderlabs/rudder-sdk-js/817)

### DIFF
--- a/src/integrations/Braze/browser.js
+++ b/src/integrations/Braze/browser.js
@@ -6,7 +6,6 @@ import { del } from "obj-case";
 import Logger from "../../utils/logger";
 import { LOAD_ORIGIN } from "../ScriptLoader";
 import { BrazeOperationString, NAME } from "./constants";
-import { isDefinedAndNotNull } from "../../utils/utils";
 
 const logger = new Logger(NAME);
 /*
@@ -18,8 +17,7 @@ class Braze {
     if (analytics.logLevel) logger.setLogLevel(analytics.logLevel);
     this.appKey = config.appKey;
     this.trackAnonymousUser = config.trackAnonymousUser;
-    this.enableBrazeLogging = isDefinedAndNotNull(config.enableBrazeLogging) ?
-        config.enableBrazeLogging : false;
+    this.enableBrazeLogging = config.enableBrazeLogging || false;
     this.enableHtmlInAppMessages = config.enableHtmlInAppMessages || false;
     this.allowUserSuppliedJavascript =
       config.allowUserSuppliedJavascript || false;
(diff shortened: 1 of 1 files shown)


src/integrations/Braze/browser.js
```diff
@@ -96,9 +97,8 @@ class Braze {
     })(window, document, "script");
 
     window.braze.initialize(this.appKey, {
-      enableLogging: true,
+      enableLogging: this.enableLoggingToConsole,
       baseUrl: this.endPoint,
-      enableHtmlInAppMessages: this.enableHtmlInAppMessages,
```
sandeepdsvs, line 101: What is this `enableHtmlInAppMessages` and why are we removing it ?

sandeepdsvs (dismissed):

src/integrations/Braze/browser.js
sandeepdsvs: Add the details in the description

src/integrations/Braze/browser.js
```diff
@@ -96,9 +97,8 @@ class Braze {
     })(window, document, "script");
 
     window.braze.initialize(this.appKey, {
-      enableLogging: true,
+      enableLogging: this.enableLoggingToConsole,
```
ItsSudip, line 100:
```suggestion
      enableLogging: this.enableLoggingToConsole || true,
```

ItsSudip (changes requested):

src/integrations/Braze/browser.js
aashishmalik: by default, this value will be false as it's boolean

ItsSudip: This destination is already used and an old one. If you make it false using a config variable then it will not be reflected in all the customers. So this variable will not be available for old customers.

aashishmalik: I think it's better to not show braze logs to end user. will update as you suggested

ItsSudip: Make sense. Verify one thing, if you are sending undefined in place of `enableLogging`, is the SDK handling this or not?

aashishmalik: updated

src/integrations/Braze/browser.js
```diff
@@ -17,6 +18,8 @@ class Braze {
     if (analytics.logLevel) logger.setLogLevel(analytics.logLevel);
     this.appKey = config.appKey;
     this.trackAnonymousUser = config.trackAnonymousUser;
+    this.enableBrazeLogging = isDefinedAndNotNull(config.enableBrazeLogging) ?
+        config.enableBrazeLogging : false;
```
ItsSudip, line 22:
```suggestion
    this.enableBrazeLogging = config.enableBrazeLogging || false;
```

ItsSudip (approved):

sandeepdsvs (approved):

PR #822, opened 2023-01-04
fix: ga4- fix total to price for events

yashasvibajpai:
## PR Description

We are fixing sending price for supported events instead of total.
- product viewed
- product added
- product removed
- add to wishlist

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security

### DIFF
--- a/src/integrations/GA4/ECommerceEventConfig.js
+++ b/src/integrations/GA4/ECommerceEventConfig.js
@@ -169,7 +169,7 @@ const eventNamesConfigArray = [
     hasItem: true,
     includeList: [
       eventParametersConfigArray.Currency,
-      eventParametersConfigArray.Total,
+      eventParametersConfigArray.Price,
     ],
   },
   {
@@ -182,7 +182,7 @@ const eventNamesConfigArray = [
     hasItem: true,
     includeList: [
       eventParametersConfigArray.Currency,
-      eventParametersConfigArray.Total,
+      eventParametersConfigArray.Price,
     ],
   },
   {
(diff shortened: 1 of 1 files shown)


shrouti1507 (dismissed):

src/integrations/GA4/ECommerceEventConfig.js
```diff
@@ -169,7 +169,7 @@ const eventNamesConfigArray = [
     hasItem: true,
     includeList: [
       eventParametersConfigArray.Currency,
-      eventParametersConfigArray.Total,
+      eventParametersConfigArray.Price,
```
ItsSudip, line 172: It looks like a breaking change. Can't we support both the total and price?
```suggestion
      eventParametersConfigArray.Total,
      eventParametersConfigArray.Price,
```

yashasvibajpai: We could, but that essentially means we are deviating from the rudderstack ecommerce spec.
For example, in this event (Product Viewed), we do not support a total property in the root ([reference](https://www.rudderstack.com/docs/event-spec/ecommerce-events-spec/ordering/#product-viewed)) and rather support a `price` property. 
We could though keep it for backward compatibility purposes, if this is okay.

ItsSudip: Let's keep it backword compatible.

yashasvibajpai: Done

ItsSudip (approved):

shrouti1507 (approved):

sandeepdsvs (approved):

PR #829, opened 2023-01-10
fix: ga4 cookie issue

mihir-4116:
## PR Description

- problem : if we clear the cookie and then load sdk and then fire track call, clientId is not going. we need to refresh it again in order it to work. so while debugging i found out that the function where we are getting clientId and sessionId is being called before clientId cookie is set by GA4 so that's why we are not getting it

- solution : when GA4 integration is ready and loaded at that time we are not getting the cookie value which we requires. one possible solution to this is prepare `integrationsDataObj` on every call. this way we are making sure that we always have correct clientId and sessionId for GA4 

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


saikumarrs (changes requested):
Please add explicit documentation for the changes introduced.

mihir-4116:
> documentation link added

src/integrations/GA4/browser.js
```diff
@@ -75,11 +88,11 @@ export default class GA4 {
 
   /* utility functions ---Start here ---  */
   isLoaded() {
-    return !!(window.gtag && window.gtag.push !== Array.prototype.push);
+    return !!(this.clientId && this.sessionId);
   }
 
   isReady() {
-    return !!(window.gtag && window.gtag.push !== Array.prototype.push);
+    return this.isLoaded();
```
saikumarrs, line 95: Add documentation for this logic too.

saikumarrs (dismissed):

src/integrations/GA4/browser.js
mihir-4116: updated

src/integrations/GA4/browser.js
```diff
@@ -74,13 +87,18 @@ export default class GA4 {
   }
 
   /* utility functions ---Start here ---  */
+
+  /**
+   * we are making sure that for every user clientId and sessionId exists by calling window.gtag function
+   */
```
saikumarrs, line 93:
```suggestion

  /**
   * If the gtag is successfully initialized, client ID and session ID fields will have valid values for the given GA4 configuration
   */
```

mihir-4116: updated

saikumarrs (approved):

src/integrations/GA4/browser.js
```diff
@@ -74,13 +87,18 @@ export default class GA4 {
   }
 
   /* utility functions ---Start here ---  */
+
+  /**
+   * If the gtag is successfully initialized, client ID and session ID fields will have valid values for the given GA4 configuration
+   */
   isLoaded() {
-    return !!(window.gtag && window.gtag.push !== Array.prototype.push);
+    return !!(this.clientId && this.sessionId);
```
krishna2020, line 95: Do we need double logical not !!

mihir-4116: `!!(this.clientId && this.sessionId);` and `this.clientId && this.sessionId` are exectly same but since we are using` !!` across sdk that's why we kept it

krishna2020 (approved):

ItsSudip (approved):

MoumitaM (approved):

PR #888, opened 2023-02-08
fix: googleads config order

mihir-4116:
## PR Description

- for any event in google ads, we will check first for the new config values. if new config values are present then we will proceed with that else we will fall back to old config

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security

### DIFF
--- a/src/integrations/GoogleAds/browser.js
+++ b/src/integrations/GoogleAds/browser.js
@@ -26,8 +26,8 @@ class GoogleAds {
     this.sendPageView = config.sendPageView || true;
     this.conversionLinker = config.conversionLinker || true;
     this.disableAdPersonalization = config.disableAdPersonalization || false;
-    this.trackConversions = config.trackConversions || false;
-    this.trackDynamicRemarketing = config.trackDynamicRemarketing || false;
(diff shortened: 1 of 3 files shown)


src/integrations/GoogleAds/browser.js
```diff
@@ -26,8 +26,8 @@ class GoogleAds {
     this.sendPageView = config.sendPageView || true;
     this.conversionLinker = config.conversionLinker || true;
     this.disableAdPersonalization = config.disableAdPersonalization || false;
-    this.trackConversions = config.trackConversions || false;
-    this.trackDynamicRemarketing = config.trackDynamicRemarketing || false;
+    this.trackConversions = config.trackConversions;
+    this.trackDynamicRemarketing = config.trackDynamicRemarketing;
```
ItsSudip, line 30: Do we need to remove the default values?

utsabc: Yes otherwise we cant apply `isDefinedNotNull` check

utsabc (approved):

ItsSudip (approved):

PR #890, opened 2023-02-08
fix: added logic to capture full url for ga4 page_location param

mihir-4116:
## PR Description

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security

### DIFF
--- a/src/integrations/GA4/utils.js
+++ b/src/integrations/GA4/utils.js
@@ -173,8 +173,6 @@ function getDestinationEventProperties(
     props,
     contextOp
   );
-  // overriding js sdk page param to capture full url
-  propsWithCustomFields.page_location = window.location.href;
   return propsWithCustomFields;
 }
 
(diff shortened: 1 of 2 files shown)


src/integrations/GA4/utils.js
```diff
@@ -173,6 +173,8 @@ function getDestinationEventProperties(
     props,
     contextOp
   );
+  // overriding js sdk page param to capture full url
+  propsWithCustomFields.page_location = window.location.href;
```
utsabc, line 177: Can we not update in the mapping part?

mihir-4116: @utsabc this is basically coming from analytics page call so we need to override it here

utsabc: But where are we mapping in `propsWithCustomFields` prior to this?

mihir-4116: updated as we discussed

utsabc (approved):

utsabc:
Add some unit test case for this need not be exhaustive

koladilip (approved):

PR #906, opened 2023-02-15
feat(ga4): add sent_to parameter in gtag event call

Gauravudia:
## PR Description

- Adding the `send_to` parameter in gtag event call to ensure that event will only be sent to the GA4 property with the measurement id configured in webapp.
- send_to parameter helps in routing data to a patricular GA4 property when more than one instance/properties are loaded.

## Notion ticket

[Ticket link](https://www.notion.so/rudderstacks/GA4-Send-send_to-parameter-with-gtag-during-event-calls-d4b077c737004a5f86b1fd22eb756695)

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security

### DIFF
--- a/src/integrations/GA4/browser.js
+++ b/src/integrations/GA4/browser.js
@@ -1,6 +1,6 @@
 /* eslint-disable class-methods-use-this */
-import ScriptLoader from "../ScriptLoader";
-import Logger from "../../utils/logger";
+import ScriptLoader from '../ScriptLoader';
+import Logger from '../../utils/logger';
 
 import {
   isReservedName,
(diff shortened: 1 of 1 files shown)


mihir-4116 (approved):

src/integrations/GA4/browser.js
```diff
@@ -160,7 +161,11 @@ export default class GA4 {
     if (checkRequiredParameters && !hasRequiredParameters(parameters, eventMappingObj)) {
       throw Error('Payload must have required parameters..');
     }
-    window.gtag("event", event, parameters);
+    const params = { ...parameters };
+    if (this.addSendToParameter) {
+      params.send_to = this.measurementId;
+    }
+    window.gtag('event', event, params);
```
ItsSudip, line 168:
```suggestion
    if (this.addSendToParameter) {
      parameters.send_to = this.measurementId;
    }
    window.gtag('event', event, parameters);
```
Do we need to create another variable?

Gauravudia: ESlint error because we are modifying functional argument. So i created a new variable

ItsSudip (approved):

PR #990, opened 2023-04-05
feat: config manager

MoumitaM:
## PR Description
Config Manager:
1. Validate and process load options
2. Save the options in the global state
3. Fetch source config
4. Construct the values that we want to retain in the state from the sourceConfig response and save source, destination, lifecycle, and reporting-related information in a global state.

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


packages/analytics-js/src/components/configManager/types.ts
```diff
@@ -81,6 +85,13 @@ export type SourceConfigResponse = {
     createdAt: string;
     createdBy: string;
     deleted: boolean;
-    dataplanes?: Record<ResidencyServerRegion, RegionDetails>;
+    dataplanes?: Record<ResidencyServerRegion, RegionDetails[]>;
   };
 };
+
+export interface IConfigManager {
```
bardisg, line 92: please add all methods that the ConfigManager Class has in this inteface

packages/analytics-js/src/components/configManager/util/commonUtil.ts
```diff
@@ -0,0 +1,34 @@
+import { Nullable } from '@rudderstack/analytics-js/types';
+
+/**
+ * Removes trailing slash from url
+ * @param url
+ * @returns url
+ */
+const removeTrailingSlashes = (url: string | null): Nullable<string> =>
```
bardisg, line 8: please move this one to components utilities string file

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,150 @@
...
+
+    // set the final load option in state
+    batch(() => {
+      // set application lifecycle state in state
+      if (state.loadOptions.value.logLevel) {
+        state.lifecycle.logLevel.value = state.loadOptions.value.logLevel;
+      }
+      state.lifecycle.integrationsCDNPath.value = intgCdnUrl;
+      if (state.loadOptions.value.configUrl) {
+        // TODO: pass the load option for lockedIntegrations version as query param in the future for metrics capturing
```
bardisg, line 58: lets add this todo also please

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,150 @@
...
 
-  getConfig() {
-    // Once done call all ready
-    this.onReady();
+  /**
+   * A callback function that is executed once we fetch the source config response.
+   * Use to construct and store information that are dependent on the sourceConfig.
+   * @param res source config response
+   */
+  processResponse(res: SourceConfigResponse | string | undefined) {
```
bardisg, line 71: res?: SourceConfigResponse | string

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,150 @@
...
+    if (!res || typeof res === 'string') {
```
bardisg, line 72: if it shoudl always be object better to check that if type is not object instead of string

bardisg (changes requested):

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
```diff
@@ -0,0 +1,63 @@
...
+};
+/**
+ * A function to determine the dataPlaneUrl
+ * @param {Object} dataPlaneUrls An object containing dataPlaneUrl for different region
+ * @returns string
+ */
+const resolveDataPlaneUrl = (
+  dataplanes:
+    | Record<
+        ResidencyServerRegion | ResidencyServerRegion.US | ResidencyServerRegion.EU,
```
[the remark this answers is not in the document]
MoumitaM: If the type is Record<ResidencyServerRegion, RegionDetails[]> and we provide either US/EU then it is not working.

packages/analytics-js/src/components/configManager/util/commonUtil.ts
```diff
@@ -0,0 +1,34 @@
...
+  url && url.endsWith('/') ? url.replace(/\/+$/, '') : url;
+
+/**
+ * Determines if staging SDK is being used
```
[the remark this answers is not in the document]
MoumitaM: It is to calculate the path of integration SDKs.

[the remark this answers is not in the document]
MoumitaM: In the lifecycle state, there is a parameter `isStaging`. Should we omit it then?

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Would like to understand the need for this.

packages/analytics-js/src/state/types.ts
```diff
@@ -276,9 +276,21 @@ export type RegionDetails = {
...
+  config: SourceConfig;
+  name?: string;
+  writeKey?: string;
+  workspaceId?: string;
+  enabled?: boolean;
+  dataplanes?: Record<ResidencyServerRegion, RegionDetails[]>;
+};
+
+export type SourceConfig = {
+  statsCollection?: {
```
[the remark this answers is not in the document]
MoumitaM: Done

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
```diff
@@ -0,0 +1,63 @@
+import { RegionDetails, ResidencyServerRegion } from '@rudderstack/analytics-js/state/types';
+import { isValidUrl } from './validate';
+
+const DEFAULT_REGION = 'US';
+
+export type RegionUrl = {
```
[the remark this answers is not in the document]
MoumitaM: Done

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @param {Object} dataPlaneUrls An object containing dataPlaneUrl for different region
+ * @returns string
+ */
+const resolveDataPlaneUrl = (
+  dataplanes:
+    | Record<
+        ResidencyServerRegion | ResidencyServerRegion.US | ResidencyServerRegion.EU,
+        RegionDetails[]
+      >
+    | undefined,
```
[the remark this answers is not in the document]
MoumitaM: Done

packages/analytics-js/src/components/configManager/types.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/util/commonUtil.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/util/commonUtil.ts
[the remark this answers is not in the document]
MoumitaM: For now we will keep this

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,148 @@
...
+   * A function to validate, construct and store loadOption, lifecycle, source and destination
+   * config related information in global state
+   */
   init() {
-    // TODO: get writekey, dataplaneUrl and loadOptions from state to use
+    validateLoadArgs(state.lifecycle.writeKey.value, state.lifecycle.dataPlaneUrl.value);
+    // determine the sourceConfig url
+    const intgCdnUrl = getIntegrationsCDNPath(
+      'process.package_version',
+      state.loadOptions.value.lockIntegrationsVersion as boolean,
```
saikumarrs, line 42: This typecasting shouldn't happen here. They have to be pre-validated.

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,148 @@
...
+    // determine if the staging SDK is being used
+    // TODO: deprecate this in new version and stop adding '-staging' in filenames
+    const { isStaging } = getSDKUrlInfo();
+
+    // set the final load option in state
+    batch(() => {
+      // set application lifecycle state in state
+      if (state.loadOptions.value.logLevel) {
+        state.lifecycle.logLevel.value = state.loadOptions.value.logLevel;
+      }
```
saikumarrs, line 55: Do we have to validate? It is already initialized with a default value, right?

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,148 @@
...
 
-  onReady() {
-    state.lifecycle.status.value = 'configured';
+  /**
+   * A function to fetch source config either from /sourceConfig endpoint
+   * or from getSourceConfig load option
+   * @returns
+   */
+  getConfig() {
+    const sourceConfigOption = state.loadOptions.value.getSourceConfig;
```
saikumarrs, line 110: `sourceConfigFunc` is a better name than `sourceConfigOption`

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,148 @@
...
+    if (sourceConfigOption) {
+      // fetch source config from the function
+      const res = sourceConfigOption();
```
saikumarrs, line 113: Is it already validated to be of function type?

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,148 @@
...
+      }
+      return;
+    }
+
+    // fetch source config from config url API
+    this.httpClient.getAsyncData({
+      url: state.lifecycle.sourceConfigUrl.value,
+      options: {
+        headers: {
+          'Content-Type': undefined,
```
saikumarrs, line 134: Use `Accept` header instead with `application/json` value.

packages/analytics-js/src/services/HttpClient/xhr/xhrRequestHandler.ts
```diff
@@ -80,7 +80,9 @@ const xhrRequest = (
     xhr.open(options.method, options.url);
 
     Object.keys(options.headers).forEach(headerName => {
```
saikumarrs, line 82: Headers are optional. We should validate for that.

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,148 @@
...
-  getConfig() {
-    // Once done call all ready
-    this.onReady();
+  /**
+   * A callback function that is executed once we fetch the source config response.
+   * Use to construct and store information that are dependent on the sourceConfig.
+   * @param res source config response
+   */
+  processConfig(res?: SourceConfigResponse | string) {
+    if (!res || typeof res !== 'object') {
```
saikumarrs, line 71: The above definition supports `string`.

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,148 @@
...
+      throw Error('Unable to fetch source config');
+    }
+    // determine the dataPlane url
+    const dataPlaneUrl = resolveDataPlaneUrl(
+      res.source.dataplanes,
```
saikumarrs, line 76: Don't we have to make sure `source` is defined?

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,148 @@
...
+      // set device mode destination related information in state
+      state.destinations.value = nativeDestinations;
+
+      // set application lifecycle state
+      state.lifecycle.activeDataplaneUrl.value = dataPlaneUrl;
+      state.lifecycle.status.value = 'configured';
+
+      // set the values in state for reporting slice
+      state.reporting.isErrorReportingEnabled.value =
+        res.source.config.statsCollection.errorReports.enabled || false;
```
saikumarrs, line 98: The key path is now `statsCollection.errors.enabled`.

saikumarrs (commented):
Partial review 1 - reviewed only source code

packages/analytics-js/src/state/slices/loadOptions.ts
```diff
@@ -24,6 +24,10 @@ const defaultLoadOptions: LoadOptions = {
   },
   sameSiteCookie: 'Lax',
   polyfillIfRequired: true,
+  integrations: { All: true },
+  useBeacon: false,
+  lockIntegrationsVersion: false,
+  uaChTrackLevel: 'none',
```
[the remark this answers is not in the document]
MoumitaM: This file only has the default load options. Didn't get your point.

packages/analytics-js/src/components/configManager/util/validate.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const validateDataPlaneUrl = (dataPlaneUrl: string) => {
+  if (dataPlaneUrl.trim().length === 0) {
+    throw Error('Unable to load the SDK due to invalid dataPlaneUrl');
+  }
+};
+
+const validateLoadArgs = (writeKey?: string, dataPlaneUrl?: string) => {
+  validateWriteKey(writeKey);
+  if (dataPlaneUrl) {
```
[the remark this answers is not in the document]
MoumitaM: Is there any benefit moving the if condition?

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @returns string
+ */
+const resolveDataPlaneUrl = (
+  dataplanes?: Record<ResidencyServerRegion, RegionDetails[]>,
+  serverUrl?: string,
+  residencyServer?: ResidencyServerRegion,
+) => {
+  // Check if dataPlanes object is present in source config
+  if (dataplanes && Object.keys(dataplanes).length > 0) {
+    const region = residencyServer || DEFAULT_REGION;
```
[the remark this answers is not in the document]
MoumitaM: Now the residency server input is an enum. User can not provide anything other the supported values.

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
```diff
@@ -0,0 +1,53 @@
...
+    const regionUrlArr: RegionDetails[] = dataplanes[region] || dataplanes[DEFAULT_REGION];
+
+    if (regionUrlArr) {
```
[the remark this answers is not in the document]
MoumitaM: Didn't get your point. Please explain

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
```diff
@@ -0,0 +1,53 @@
...
+      const defaultUrl = getDefaultUrlofRegion(regionUrlArr);
+      if (defaultUrl) {
+        return defaultUrl;
+      }
+    }
+  }
+  // return the dataPlaneUrl provided in load API(if available)
+  if (serverUrl) {
```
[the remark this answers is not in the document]
MoumitaM: We are not validating here, just checking if it exists.

packages/analytics-js/src/components/configManager/util/validate.ts
```diff
@@ -0,0 +1,24 @@
+const WRITE_KEY_LENGTH = 27;
+
+const validateWriteKey = (writeKey?: string) => {
+  if (!writeKey || writeKey.trim().length !== WRITE_KEY_LENGTH) {
+    throw Error('Unable to load the SDK due to invalid writeKey');
+  }
+};
+
+const isValidUrl = (url: string): boolean => url.trim().length > 0;
```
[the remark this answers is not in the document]
MoumitaM: `new URL` is not supported in IE 11, Safari 7. In that case, we have to use polyfill.

packages/analytics-js/src/components/configManager/util/commonUtil.ts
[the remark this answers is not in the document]
bardisg: we will remove as soon as we prepare the ci for v3 correctly

packages/analytics-js/src/services/HttpClient/xhr/xhrRequestHandler.ts
bardisg: headers is not optional, if none were passes sensible defaults are merged with them DEFAULT_XHR_REQUEST_OPTIONS

packages/analytics-js/src/components/configManager/ConfigManager.ts
bardisg: getSourceConfig?: () => string | apiObject | Promise<apiObject> | Promise<string> 

is what is supported by v1.1. types.

Was this not correct in v1.1?

packages/analytics-js/src/services/HttpClient/xhr/xhrRequestHandler.ts
saikumarrs: Makes sense. I'm talking about the input headers to this client.

packages/analytics-js/src/components/configManager/util/validate.ts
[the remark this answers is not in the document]
saikumarrs: By definition, it's a validation method. Why not move this also inside?

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
saikumarrs: What about CDN users?

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
saikumarrs: Array.isArray

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
saikumarrs: The if condition.

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
MoumitaM: Done

packages/analytics-js/src/state/types.ts
```diff
@@ -278,9 +278,21 @@ export type RegionDetails = {
...
-  name: string;
-  writeKey: string;
-  workspaceId: string;
-  enabled: boolean;
-  dataplanes: Record<ResidencyServerRegion, RegionDetails>;
+  config?: SourceConfig;
+  name?: string;
+  writeKey?: string;
+  workspaceId?: string;
+  enabled?: boolean;
```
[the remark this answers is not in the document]
saikumarrs: Yes.

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
MoumitaM: dataPlaneUrl is now optional. We only validate it if it is present.
If dataPlaneUrl is not provided as the load call argument we need the if condition here.

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
```diff
@@ -0,0 +1,53 @@
...
+    return serverUrl;
+  }
+
+  // TODO: do we need to format the error message?
+  // Throw error if correct data plane url is not provided
+  throw Error('Unable to load the SDK due to invalid data plane url');
```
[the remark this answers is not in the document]
MoumitaM: Do you mean the dataPlaneUrl in the load call?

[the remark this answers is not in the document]
saikumarrs: Yes, everything that we're validating here.

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
saikumarrs: Aren't we checking if the URL provided is a valid URL elsewhere?

packages/analytics-js/src/components/configManager/ConfigManager.ts
saikumarrs: It's better to only accept Boolean values for the input. Hence, we can validate them accordingly.

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,148 @@
...
+  logger?: ILogger;
+  hasErrorHandler = false;
+  hasLogger = false;
+
+  constructor(httpClient: IHttpClient, errorHandler?: IErrorHandler, logger?: ILogger) {
+    this.errorHandler = errorHandler;
+    this.logger = logger;
+    this.httpClient = httpClient;
+    this.hasErrorHandler = Boolean(this.errorHandler);
+    this.hasLogger = Boolean(this.logger);
```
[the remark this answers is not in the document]
saikumarrs: What use do these serve? Apologies, I couldn't find the code that are using these booleans.

packages/analytics-js/src/components/configManager/ConfigManager.ts
saikumarrs: Yes, only functions are allowed but we have to validate if it's indeed a function before trying to invoke it.

packages/analytics-js/src/components/configManager/ConfigManager.ts
saikumarrs: In that case, this can be removed from here.

packages/analytics-js/src/components/configManager/ConfigManager.ts
saikumarrs: That's accurate. The logic below assumes res to be an object. However from the function's definition, we also support string and here the validation is only done for object type.

packages/analytics-js/src/components/configManager/ConfigManager.ts
saikumarrs: Yes.

packages/analytics-js/src/components/configManager/ConfigManager.ts
saikumarrs: This would also work in our advantage where the impact radius would be minimal in case something crazy happens after the release.
Moreover, once we pick the customers who are our initial test users, we can ask them to upgrade to the version with Bugsnag feature.

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Even though it is initialized with default value we need to validate here. Else it complains.

packages/analytics-js/src/components/configManager/ConfigManager.ts
bardisg: we need to unset the default Content-Type header value for thisAPI endpoint

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Yes, this is correct. Should we also validate `res` if it is a `string` and parse it?

packages/analytics-js/src/state/types.ts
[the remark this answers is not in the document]
MoumitaM: Done

packages/analytics-js/src/state/slices/loadOptions.ts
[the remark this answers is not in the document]
MoumitaM: Done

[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/validate.ts
```diff
@@ -0,0 +1,24 @@
+const WRITE_KEY_LENGTH = 27;
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/validate.ts
```diff
@@ -0,0 +1,24 @@
...
+
+const validateWriteKey = (writeKey?: string) => {
+  if (!writeKey || writeKey.trim().length !== WRITE_KEY_LENGTH) {
+    throw Error('Unable to load the SDK due to invalid writeKey');
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/validate.ts
```diff
@@ -0,0 +1,24 @@
...
+  }
+};
+
+const isValidUrl = (url: string): boolean => url.trim().length > 0;
+
+const validateDataPlaneUrl = (dataPlaneUrl: string) => {
+  if (dataPlaneUrl.trim().length === 0) {
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/validate.ts
```diff
@@ -0,0 +1,24 @@
...
+    throw Error('Unable to load the SDK due to invalid dataPlaneUrl');
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/validate.ts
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/filterDestinations.ts
```diff
@@ -0,0 +1,26 @@
...
+  destinations: ConfigResponseDestinationItem[],
+): Destination[] | [] => {
+  const nativeDestinations: Destination[] = [];
+  destinations.forEach((destination: ConfigResponseDestinationItem) => {
+    if (destination.enabled && destination.deleted !== true) {
+      nativeDestinations.push({
+        id: destination.id,
+        name: destination.destinationDefinition.name,
+        config: destination.config,
+        areTransformationsConnected: destination.areTransformationsConnected || false,
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/filterDestinations.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+const filterEnabledDestination = (
+  destinations: ConfigResponseDestinationItem[],
+): Destination[] | [] => {
+  const nativeDestinations: Destination[] = [];
+  destinations.forEach((destination: ConfigResponseDestinationItem) => {
+    if (destination.enabled && destination.deleted !== true) {
+      nativeDestinations.push({
+        id: destination.id,
+        name: destination.destinationDefinition.name,
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+/**
+ * A function to determine the dataPlaneUrl
+ * @param {Object} dataPlaneUrls An object containing dataPlaneUrl for different region
+ * @returns string
+ */
+const resolveDataPlaneUrl = (
+  dataplanes?: Record<ResidencyServerRegion, RegionDetails[]>,
+  serverUrl?: string,
+  residencyServer?: ResidencyServerRegion,
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
```diff
@@ -0,0 +1,53 @@
...
+    if (obj && isValidUrl(obj.url)) {
+      return obj.url;
+    }
+  }
+  return url;
+};
+/**
+ * A function to determine the dataPlaneUrl
+ * @param {Object} dataPlaneUrls An object containing dataPlaneUrl for different region
+ * @returns string
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
MoumitaM: Yes we are checking, but we do nothing in case it is not provided

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
MoumitaM: It will only reach this line if dataPlaneUrl is not there and doesn't make sense to add undefined

packages/analytics-js/src/components/configManager/util/validate.ts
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/commonUtil.ts
```diff
@@ -0,0 +1,26 @@
+import { removeTrailingSlashes } from '../../utilities/url';
+
+/**
+ * Determines if staging SDK is being used
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/cdnPaths.ts
```diff
@@ -0,0 +1,51 @@
...
+  customIntegrationsCDNPath: string | undefined,
+): string => {
+  let integrationsCDNPath = '';
+
+  // Get the CDN base URL from the user provided URL if any
+  if (customIntegrationsCDNPath) {
+    integrationsCDNPath = removeTrailingSlashes(customIntegrationsCDNPath) as string;
+
+    // TODO: add proper url validation
+    if (!integrationsCDNPath) {
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/util/cdnPaths.ts
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * A function that determines integration SDK loading path
+ * @param currentVersion
+ * @param lockIntegrationsVersion
+ * @param customIntegrationsCDNPath
+ * @returns
+ */
+const getIntegrationsCDNPath = (
+  currentVersion: string,
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,148 @@
...
+  }
+
+  /**
+   * A function to validate, construct and store loadOption, lifecycle, source and destination
+   * config related information in global state
+   */
   init() {
-    // TODO: get writekey, dataplaneUrl and loadOptions from state to use
+    validateLoadArgs(state.lifecycle.writeKey.value, state.lifecycle.dataPlaneUrl.value);
+    // determine the sourceConfig url
```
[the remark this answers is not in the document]
MoumitaM: done

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: done

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: done

packages/analytics-js/src/components/configManager/ConfigManager.ts
saikumarrs: Yes. Otherwise, it could be a breaking change.

packages/analytics-js/src/components/configManager/ConfigManager.ts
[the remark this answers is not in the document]
MoumitaM: We haven't used this in this module yet

packages/analytics-js/src/components/utilities/url.ts
```diff
@@ -8,4 +8,18 @@ import { Nullable } from '@rudderstack/analytics-js/types';
 const removeTrailingSlashes = (url: string | null): Nullable<string> =>
   url && url.endsWith('/') ? url.replace(/\/+$/, '') : url;
 
-export { removeTrailingSlashes };
+/**
+ * Checks if provided url is valid or not
+ * @param url
+ * @returns boolean
```
saikumarrs, line 14:
```suggestion
 * @returns true if `url` is valid and false otherwise
```

packages/analytics-js/src/components/configManager/util/validate.ts
```diff
@@ -1,24 +1,22 @@
-const WRITE_KEY_LENGTH = 27;
+import { isValidUrl } from '../../utilities/url';
+
+const WRITE_KEY_LENGTH = 27; // writeKey generated in control plane consists of 27 character
```
saikumarrs, line 3:
```suggestion
// The write key is a KSUID, 
// https://github.com/segmentio/ksuid
const WRITE_KEY_LENGTH = 27;
```

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
```diff
@@ -21,23 +21,23 @@ const getDefaultUrlofRegion = (urls: RegionDetails[]) => {
 /**
  * A function to determine the dataPlaneUrl
  * @param {Object} dataPlaneUrls An object containing dataPlaneUrl for different region
+ * @param {String} serverUrl dataPlaneUrl provided in the load call
+ * @param {String} residencyServerRegion User provided residency server region
  * @returns string
```
saikumarrs, line 26:
```suggestion
 * @returns The data plane URL string to use
```

packages/analytics-js/src/components/configManager/util/cdnPaths.ts
```diff
@@ -24,8 +24,7 @@ const getIntegrationsCDNPath = (
   if (customIntegrationsCDNPath) {
     integrationsCDNPath = removeTrailingSlashes(customIntegrationsCDNPath) as string;
 
-    // TODO: add proper url validation
-    if (!integrationsCDNPath) {
+    if (!integrationsCDNPath || (integrationsCDNPath && !isValidUrl(integrationsCDNPath))) {
       const errorMsg = 'CDN base URL for integrations is not valid';
 
       throw Error(`Failed to load Rudder SDK: ${errorMsg}`);
```
saikumarrs, line 30:
```suggestion
      throw Error(`Failed to load RudderStack SDK: ${errorMsg}`);
```

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -36,26 +36,26 @@ class ConfigManager implements IConfigManager {
    */
   init() {
     validateLoadArgs(state.lifecycle.writeKey.value, state.lifecycle.dataPlaneUrl.value);
-    // determine the sourceConfig url
+    const lockIntegrationsVersion = state.loadOptions.value.lockIntegrationsVersion || false;
```
saikumarrs, line 39:
```suggestion
    const lockIntegrationsVersion = state.loadOptions.value.lockIntegrationsVersion === true;
```

packages/analytics-js/src/components/configManager/util/commonUtil.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @returns { sdkURL, isStaging }
+ */
+const getSDKUrlInfo = () => {
+  const scripts = document.getElementsByTagName('script');
+  let sdkURL;
+  let isStaging = false;
+  // eslint-disable-next-line unicorn/no-for-loop
```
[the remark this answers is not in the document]
MoumitaM: We can not use Array.some on scripts.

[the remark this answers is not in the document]
saikumarrs: https://stackoverflow.com/questions/19324700/how-to-loop-through-all-the-elements-returned-from-getelementsbytagname

packages/analytics-js/__mocks__/fixtures.ts
```diff
@@ -66,6 +66,61 @@ const dummyInitOptions = {
   enable: true,
 };
 
+const dummySourceConfigResponse = {
```
saikumarrs, line 69: Can you make sure we're not using any actual IDs here?

packages/analytics-js/__mocks__/msw.handlers.ts
```diff
@@ -53,6 +53,9 @@ const handlers = [
       ctx.body(fileContents),
     );
   }),
+  rest.get(`${dummyDataplaneHost}/sourceConfig*`, (req, res, ctx) => {
```
saikumarrs, line 56: Why `*` at the end of the URL?

packages/analytics-js/__mocks__/msw.handlers.ts
```diff
@@ -1,5 +1,5 @@
 import { rest } from 'msw';
-import { dummyDataplaneHost } from './fixtures';
+import { dummyDataplaneHost, dummySourceConfigResponse } from './fixtures';
 
 // TODO: Why on data plane we allow-origin the domain but in sourceConfig is wildcard?
```
saikumarrs, line 4: @bardisg This should not be the case. Do you see otherwise?

packages/analytics-js/__mocks__/fixtures.ts
```diff
@@ -66,6 +66,61 @@ const dummyInitOptions = {
...
+        name: 'GA4 for JS SDK',
+        updatedAt: '2023-03-14T11:34:29.216Z',
+        enabled: true,
+        deleted: false,
+        destinationDefinition: {
+          name: 'GA4',
+          displayName: 'Google Analytics 4 (GA4)',
+          updatedAt: '2023-03-14T11:21:29.656Z',
+        },
+        areTransformationsConnected: true,
```
saikumarrs, line 118: Do we want this to be true?

packages/analytics-js/__tests__/components/configManager/ConfigManager.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    server.close();
+  });
+
+  it('should throw an error for invalid writeKey', () => {
+    state.lifecycle.writeKey.value = 'INVALID-WRITE-KEY';
+    try {
+      configManagerInstance.init();
+    } catch (e) {
+      expect(e.message).toBe(errorMsg);
+    }
```
saikumarrs, line 85: The correct way to test exception-throwing code.
```suggestion
    expect(() => { configManagerInstance.init() }).toThrow(errorMsg);
```

packages/analytics-js/__tests__/components/configManager/ConfigManager.test.ts
```diff
@@ -0,0 +1,208 @@
...
+  });
+  it('should throw error for invalid data plane url', () => {
+    state.lifecycle.writeKey.value = sampleWriteKey;
+    state.lifecycle.dataPlaneUrl.value = ' ';
```
saikumarrs, line 89: Global data like these should be reset at the end of each test if they are mutated during the test.

packages/analytics-js/__tests__/components/configManager/validate.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  const sampleWriteKey = '2LoR1TbVG2bcISXvy7DamldfkgO';
+  const sampleDataPlaneUrl = 'https://www.dummy.url';
+  const errorMsg = 'Unable to load the SDK due to invalid write key: "INVALID-WRITE-KEY"';
+
+  it('should not throw error for valid write key', () => {
+    try {
+      validateLoadArgs(sampleWriteKey);
+    } catch (e) {
+      expect(e).not.toThrow(errorMsg);
+    }
```
saikumarrs, line 13: Incorrect.
```suggestion
   expect(() => { validateLoadArgs(sampleWriteKey) }).not.toThrow(errorMsg);
```

saikumarrs (commented):
Review complete.

packages/analytics-js/src/components/configManager/util/validate.ts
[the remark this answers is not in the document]
saikumarrs: Also, don't validate the length of the key to allow flexibility.

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,21 +1,148 @@
...
+
+    // set the final load option in state
+    batch(() => {
+      // set application lifecycle state in state
+      if (state.loadOptions.value.logLevel) {
+        state.lifecycle.logLevel.value = state.loadOptions.value.logLevel;
+      }
+      state.lifecycle.integrationsCDNPath.value = intgCdnUrl;
+      if (state.loadOptions.value.configUrl) {
+        state.lifecycle.sourceConfigUrl.value = `${state.loadOptions.value.configUrl}/sourceConfig/?p=process.module_type&v=process.package_version&writeKey=${state.lifecycle.writeKey.value}&lockIntegrationsVersion=${state.loadOptions.value.lockIntegrationsVersion}`;
```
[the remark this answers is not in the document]
MoumitaM: Adressed

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: why do we want to lock the version by default?

saikumarrs: It's quite the opposite. Read again.
We're assigning `true` only if the incoming option is `true` and everything else resolves to `false`.

MoumitaM: Got it

packages/analytics-js/__mocks__/msw.handlers.ts
bardisg: I believe that serverside shouls always allow as cross orgin the refereer URL. Having said this the sourceConfig endpoint response comes from CloudFront and by using different response headers as cache criteria would probabaly blow up cost. 

This is something to consider afteer we do the v3 MVP and if server can fix it we can potentally start using fetch instead of legacy XHR (as fetch does not allow CORS requests with credentials adn wildcard), thus the todo in the comments

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -67,10 +67,25 @@ class ConfigManager implements IConfigManager {
...
    */
-  processConfig(res?: SourceConfigResponse | string) {
-    if (!res || typeof res !== 'object') {
+  processConfig(response?: SourceConfigResponse | string) {
+    if (!response) {
       throw Error('Unable to fetch source config');
     }
+    let res: SourceConfigResponse;
+    if (typeof response === 'string') {
+      res = JSON.parse(response);
```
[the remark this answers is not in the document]
MoumitaM: Done

packages/analytics-js/__tests__/components/configManager/commonUtil.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  script.id = 'SOME_ID';
+  document.head.appendChild(script);
+};
+
+const removeScriptElement = () => {
+  const scriptElem = document.getElementById('SOME_ID');
+  scriptElem.remove();
+};
+
+describe('Common util: getSDKUrlInfo', () => {
```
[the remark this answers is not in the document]
MoumitaM: Done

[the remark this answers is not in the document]
MoumitaM: Done

packages/analytics-js/__tests__/components/configManager/ConfigManager.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    }
+  });
+  it('should throw an error for sourceConfig response in string format', () => {
+    try {
+      configManagerInstance.processConfig(JSON.stringify(dummySourceConfigResponse));
+    } catch (e) {
+      expect(e.message).toBe(errorMsgSourceConfigResponse);
+    }
+  });
+  it('should fetch the sourceConfig and call the callback with the response ', () => {
```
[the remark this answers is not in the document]
MoumitaM: Updated the description

packages/analytics-js/__mocks__/fixtures.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/util/validate.ts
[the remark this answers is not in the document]
MoumitaM: Done

packages/analytics-js/src/components/configManager/util/commonUtil.ts
```diff
@@ -0,0 +1,26 @@
...
+    if (curScriptSrc) {
+      const urlMatches = curScriptSrc.match(/^.*rudder-analytics(-staging)?(\.min)?\.js$/);
+      if (urlMatches) {
+        sdkURL = curScriptSrc;
+        isStaging = urlMatches[1] !== undefined;
+        break;
+      }
+    }
+  }
+  return { sdkURL, isStaging };
```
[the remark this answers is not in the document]
MoumitaM: Will do this improvements later phase of the project

packages/analytics-js/__mocks__/fixtures.ts
saikumarrs: The values still look real. Instead, mask them like `validLookingID123`, `dummyID` etc.

packages/analytics-js/src/components/configManager/util/commonUtil.ts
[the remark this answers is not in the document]
saikumarrs: I think we should do it now. The current code is not good enough.
At least add a TODO if we want to postpone it.
CC @bardisg

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,23 +1,168 @@
...
+    try {
+      if (typeof response === 'string') {
+        res = JSON.parse(response);
+      } else {
+        res = response;
+      }
+    } catch (e) {
+      if (this.hasErrorHandler) {
+        this.errorHandler?.onError(e, 'Config Manager: processConfig');
+      }
```
saikumarrs, line 85: The optional chaining operator is enough. No need for the `if` condition.

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,23 +1,168 @@
...
-      this.onReady();
-    }, 500);
+  /**
+   * A callback function that is executed once we fetch the source config response.
+   * Use to construct and store information that are dependent on the sourceConfig.
+   * @param res source config response
+   */
+  processConfig(response?: SourceConfigResponse | string) {
+    if (!response) {
+      throw Error('Unable to fetch source config');
```
saikumarrs, line 72: Shouldn't this be "Unable to process/parse source config"?

packages/analytics-js/src/components/configManager/ConfigManager.ts
```diff
@@ -1,23 +1,168 @@
...
+      if (typeof response === 'string') {
+        res = JSON.parse(response);
+      } else {
+        res = response;
+      }
+    } catch (e) {
+      if (this.hasErrorHandler) {
+        this.errorHandler?.onError(e, 'Config Manager: processConfig');
+      }
+      throw Error(errMessage);
```
saikumarrs, line 86: Is this error handled at upper levels? 
If no, won't this be caught by the global unhandled error and notify again?
If yes, does it handle the error again?

packages/analytics-js/src/components/configManager/util/commonUtil.ts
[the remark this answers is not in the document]
MoumitaM: Will do this improvements later phase of the project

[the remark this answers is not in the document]
saikumarrs: It's no big deal. We can do it now.

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
MoumitaM: Done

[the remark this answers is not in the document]
saikumarrs: Just logging the error isn't sufficient.

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
```diff
@@ -21,23 +22,30 @@ const getDefaultUrlofRegion = (urls: RegionDetails[]) => {
...
   serverUrl?: string,
-  residencyServer?: ResidencyServerRegion,
+  residencyServerRegion?: ResidencyServerRegion,
 ) => {
+  // validate residency server region input
+  if (
+    residencyServerRegion &&
+    !Object.values(ResidencyServerRegion).includes(residencyServerRegion)
+  ) {
+    defaultLogger.error('Invalid residencyServer input');
```
saikumarrs, line 39:
```suggestion
    defaultLogger.error(`Invalid residencyServer input: "${residencyServerRegion}"`);
```

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
saikumarrs: Also, the validation should be moved into the `if` block below.

packages/analytics-js/src/components/configManager/util/commonUtil.ts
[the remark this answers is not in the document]
MoumitaM: Done

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Added validation

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/ConfigManager.ts
saikumarrs: I can't see the updated code. Maybe it is not pushed to remote yet.

packages/analytics-js/__mocks__/fixtures.ts
MoumitaM: Done

packages/analytics-js/src/components/configManager/util/dataPlaneResolver.ts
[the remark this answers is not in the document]
MoumitaM: Done

packages/analytics-js/src/components/configManager/ConfigManager.ts
MoumitaM: try refreshing the page

packages/analytics-js/src/components/configManager/ConfigManager.ts
bardisg: The error handler will take in the error, process it and if there is logger it will log in console the appropriate message and not throw again. https://github.com/rudderlabs/rudder-sdk-js/blob/v3-staging/packages/analytics-js/src/services/ErrorHandler/ErrorHandler.ts#L58

We need to sent any errors form our components to the error handler. only if one is not defined we should plainly throw.

please see here https://github.com/rudderlabs/rudder-sdk-js/blob/v3-staging/packages/analytics-js/src/services/HttpClient/HttpClient.ts#L72 how error handler is used (cerated the onError method and use in the rest of the code in this service)

packages/analytics-js/__tests__/components/configManager/ConfigManager.test.ts
MoumitaM: Done

packages/analytics-js/__tests__/components/configManager/ConfigManager.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      defaultHttpClient,
+      defaultErrorHandler,
+      defaultLogger,
+    );
+  });
+
+  afterEach(() => {
+    server.resetHandlers();
+    server.events.removeAllListeners();
+    resetState();
```
saikumarrs, line 86: What'll be the state of these fields for the first TC? Instead, can we reset the state in beforeEach method?

packages/analytics-js/__tests__/components/configManager/ConfigManager.test.ts
```diff
@@ -123,19 +136,17 @@ describe('ConfigManager', () => {
       }),
     );
 
-    configManagerInstance.init();
+    configManagerInstance.getConfig();
 
     setTimeout(() => {
```
saikumarrs, line 141: A fixed timeout is not a proper method to ensure we are ready to verify the results. Instead, add a new property on the config manager, say `loaded`. Frequently poll that boolean and use the `done` callback to indicate test completion.

packages/analytics-js/src/components/configManager/ConfigManager.ts
saikumarrs: @MoumitaM, let's implement the same here.

packages/analytics-js/__tests__/components/configManager/ConfigManager.test.ts
bardisg: the state before each test should be the default state, unless specific values need to exist for all tests.

We should clear/reset values either on after or before.

personally I would preffer to reset state on after, prepopulate on before.

packages/analytics-js/__tests__/components/configManager/ConfigManager.test.ts
bardisg: That is not correct. We use global state with reactive abilities.

To check for a value change you can use the 
effect(() => { you expects here .... ; done()})

There is no need to poll.

saikumarrs: Using a signal would be a better solution.

packages/analytics-js/src/components/configManager/ConfigManager.ts
bardisg: yes, lets be consistent on thin in our component classes (add an onError like the HttpClient one)

MoumitaM: Done

packages/analytics-js/__tests__/components/configManager/ConfigManager.test.ts
MoumitaM: Done

MoumitaM: Done

packages/analytics-js/__tests__/components/configManager/ConfigManager.test.ts
MoumitaM: Didn't make any changes here

packages/analytics-js/src/components/utilities/url.ts
```diff
@@ -6,7 +6,7 @@ import { Nullable } from '@rudderstack/analytics-js/types';
  * @returns url
  */
 const removeTrailingSlashes = (url: string | null): Nullable<string> =>
-  url && url.endsWith('/') ? url.replace(/\/+$/, '') : url;
+  url && url.endsWith('/') ? url.substring(0, url.length - 1) : url;
```
saikumarrs, line 9: There could be more than one trailing slash. I know the previous regex resulted in a bug but try to find something different.

saikumarrs (changes requested):

packages/analytics-js/__tests__/components/configManager/ConfigManager.test.ts
```diff
@@ -124,44 +124,39 @@ describe('ConfigManager', () => {
...
     );
 
     configManagerInstance.getConfig();
 
-    setTimeout(() => {
-      expect(counter).toBe(1);
-      expect(configManagerInstance.processConfig).toHaveBeenCalled();
-    }, 1);
+    effect(() => {
+      if (counter.value === 1) {
```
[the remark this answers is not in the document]
MoumitaM: Yes, this specific test case is to check if the /sourceConfig endpoint is getting a hit or not.
"Processing the config is complete" there is another test case for that.

packages/analytics-js/src/components/utilities/url.ts
MoumitaM: Done

saikumarrs (approved):

bardisg (approved):

PR #1030, opened 2023-04-21
fix(ga4): updated hybridmode condition check

mihir-4116:
## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security

### DIFF
--- a/src/integrations/GA4/browser.js
+++ b/src/integrations/GA4/browser.js
@@ -26,7 +26,7 @@ export default class GA4 {
     this.capturePageView = config.capturePageView || 'rs';
     this.overrideSessionId = config.overrideSessionId || false;
     this.extendPageViewParams = config.extendPageViewParams || false;
-    this.isHybridModeEnabled = config.useNativeSDKToSend === false || false;
+    this.isHybridModeEnabled = config.connectionMode === 'hybrid';
   }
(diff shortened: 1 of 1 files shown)


src/integrations/GA4/browser.js
```diff
@@ -26,7 +26,7 @@ export default class GA4 {
     this.capturePageView = config.capturePageView || 'rs';
     this.overrideSessionId = config.overrideSessionId || false;
     this.extendPageViewParams = config.extendPageViewParams || false;
-    this.isHybridModeEnabled = config.useNativeSDKToSend === false || false;
+    this.isHybridModeEnabled = config.connectionMode === 'hybrid';
```
koladilip, line 29: support old and new configs

mihir-4116: we don't need that as we already migrated all existing users to new form builder so connectionMode key will always be present

ref [this](https://github.com/rudderlabs/rudder-config-backend/blob/267049e6be6b9a7ada3204e1ef62635f353ce92f/src/services/scripts.service.ts)

koladilip (approved):

sandeepdsvs (approved):

PR #1040, opened 2023-04-25
chore: add esm build to exports

lewisl9029:
## PR Description

Hi there! Looks like the exports field in `rudder-sdk-js`'s package.json didn't include the esm build for the main entry point, which kind of defeated the purpose of having an esm build in the first place 😅, so I added it here using an [export condition](https://nodejs.org/api/packages.html#conditional-exports).

Also consolidated the `./service-worker` entry point into a single entry using export conditions, but I kept the `./service-worker/index.es` entry point as an esm-only entry point for backwards compatibility for users who might have been using it previously.

That said, this change will mean users will automatically start using the esm builds for both existing entry points instead of the cjs builds, so it could still become a breaking change if there are bugs in the esm builds that don't exist in the cjs builds. May want to release this as a new major to err on the safe side.

## Notion ticket
N/A

## Screenshots
N/A

## Security


MoumitaM:
Hey @lewisl9029, Please sign the [CLA](https://rudderlabs.wufoo.com/forms/rudderlabs-contributor-license-agreement) so we can proceed with your changes.

lewisl9029:
@MoumitaM just signed! Let me know if you need anything else from my side. :)

bardisg (approved):

PR #1049, opened 2023-04-28
fix: replace rudderlabs.com with rudderstack.com for config url

saikumarrs:
## PR Description

It's recommended to use `api.rudderstack.com`.

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


MoumitaM (approved):
Changes LGTM. But won't this be a breaking change for the customers who proxy `ruddderlabs`.
Have to update our docs as well: [reference](https://www.rudderstack.com/docs/user-guides/how-to-guides/custom-domains/)

bardisg:
Also clients that use CSP policies will get issues, we could say if rudderlabs is not available then try rudderstack as a fallback

saikumarrs:
Due to concerns related to CSP, dropping these changes for now.

PR #1054, opened 2023-05-03
fix: integrations option propagation to events from load option

MoumitaM:
## PR Description

Depending on a new load option(flag) will use the integration option provided in the load call.

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


src/core/analytics.js
```diff
@@ -1133,6 +1146,10 @@ class Analytics {
       transformToRudderNames(this.loadOnlyIntegrations);
     }
 
+    if (options && options.useIntegrationsInEvents) {
```
bardisg, line 1149: rename this to useGlobalIntegrationsConfigInEvents or something so it is more explanatory.

current name useIntegrationsInEvents could be confusing as it will still sent them in events level if defined in the apiOptions

src/core/analytics.js
```diff
@@ -851,11 +852,23 @@ class Analytics {
...
-      // if not specified at event level, All: true is default
-      const clientSuppliedIntegrations = rudderElement.message.integrations || { All: true };
+      let clientSuppliedIntegrations = rudderElement.message.integrations;
+
+      if (clientSuppliedIntegrations) {
+        // structure user supplied integrations object to rudder format
+        transformToRudderNames(clientSuppliedIntegrations);
+      } else if (
+        this.useIntegrationsInEvents &&
+        Object.keys(this.loadOnlyIntegrations).length > 0
```
bardisg, line 862: if this.loadOnlyIntegrations can be null or undefined lets also check for that before Object.keys.

also better to move th check logic into a constant like shouldUseGlobalIntegrationsConfigInEvents and use the value in the confitional instead of placing the logic directly in the conditional.

easier to read, changle logic and reuse value if needed

src/core/analytics.js
```diff
@@ -1133,6 +1146,10 @@ class Analytics {
       transformToRudderNames(this.loadOnlyIntegrations);
     }
 
+    if (options && options.useIntegrationsInEvents) {
+      this.useIntegrationsInEvents = true;
+    }
```
saikumarrs, line 1151:
```suggestion
    this.useIntegrationsInEvents = options && options.useIntegrationsInEvents === true;
```

src/core/analytics.js
MoumitaM: We initialize this.loadOnlyIntegrations as {} and perform Object.assign on this. So it can never be null or undefined but still added the checking.

bardisg (approved):

saikumarrs (approved):

bardisg (approved):

PR #1088, opened 2023-05-18
feat: dataplane events queue

saikumarrs:
## PR Description

Updated dataplane and xhr queue plugins to implement retry.

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


packages/analytics-js-plugins/src/xhrQueue/index.ts
```diff
@@ -1,14 +1,156 @@
...
+import { Queue, getCurrentTimeFormatted, toBase64 } from '../utilities/common';
 
 const pluginName = PluginName.XhrQueue;
 
+let eventsQueue: Queue;
+let writeKey: string;
+let dataplaneUrl: string;
+let httpClient: IHttpClient;
+let logger: ILogger | undefined;
+let errorHandler: IErrorHandler | undefined;
```
bardisg, line 31: I am a bit worried that this will cause issues in multible instances of the sdk exist and they have multible instances of the plugin

packages/analytics-js-plugins/src/xhrQueue/index.ts
```diff
@@ -1,14 +1,156 @@
...
+      writeKey = state.lifecycle.writeKey.value as string;
+
+      httpClient = new HttpClient(errorHandler, logger);
+      httpClient.setAuthHeader(writeKey);
+
+      const finalQOpts = getNormalizedQueueOptions(
+        state.loadOptions.value.queueOptions as QueueOpts,
+      );
+
+      eventsQueue = new Queue(
```
bardisg, line 58: you can create the queue through the storage manager. there is an extension point there for that I believe

saikumarrs: I couldn't find any way to create the queue from the store/storage manager. Could you please clarify?

bardisg: I was reffering to this.pluginManager?.invokeMultiple('queuestore.create', this.setStore); of store manager but this is to create new Store not queues. 

Maybe I need to think if some refactoring is needed in localstorage-retry package to use this when creating new stores.

Unrelated to you PR, please ignore

packages/analytics-js-plugins/src/xhrQueue/index.ts
bardisg: I had in my mind that we would be passing ann required values as arguments to the extension point method call and these methods would be JS pure functions

saikumarrs: Addressed.

packages/analytics-js/src/components/pluginsManager/defaultPluginsList.ts
```diff
@@ -5,14 +5,14 @@ import { PluginName } from './types';
...
   PluginName.ConsentManager,
-  PluginName.DataplaneEventsQueue,
   PluginName.DeviceModeDestinations,
   PluginName.ErrorReporting,
   PluginName.ExternalAnonymousId,
   PluginName.GoogleLinker,
   PluginName.NativeDestinationQueue,
   PluginName.StorageEncryptionLegacy,
   PluginName.XhrQueue,
+  PluginName.BeaconQueue,
```
bardisg, line 15: this should not be in the defaults, we should add it in only if beacon is selected (and remove XhrQueue in this case).

this is done in plugin manager at getPluginsToLoadBasedOnConfig method

bardisg (approved):

bardisg (approved):

bardisg (dismissed):

package.json
```diff
@@ -20,8 +20,9 @@
     "test": "lerna run --concurrency 1 test",
     "test:ci": "lerna run --concurrency 1 test:ci",
     "check:lint": "lerna run check:lint",
-    "check:size": "lerna run check:size",
-    "check:size:build": "lerna run check:size:build",
+    "check:size": "lerna run check:size --scope \"@rudderstack/analytics-js\" --scope \"@rudderstack/analytics-js-plugins\"",
+    "check:size:build": "lerna run check:size:build --scope \"@rudderstack/analytics-js\" --scope \"@rudderstack/analytics-js-plugins\"",
+    "check:size:json": "lerna run check:size:json",
```
bardisg, line 25: why we need the --json output?

.github/workflows/build-and-quality-checks-v3.yml
```diff
@@ -43,3 +43,4 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           build_script: check:size
+          script: npm run check:size:json -- --scope '@rudderstack/analytics-js' --scope '@rudderstack/analytics-js-plugins'
```
bardisg, line 46: please change to npm run check:size , size checks are more specific and they need their builds in my branch

package.json
saikumarrs: This is for the size-limit action to forward the command to individual packages. Otherwise, it'd execute `npx size-limit --json` on the root directory and tries to compare it with the base for computing the size delta.

saikumarrs: Addressed.

.github/workflows/build-and-quality-checks-v3.yml
saikumarrs: Addressed.

bardisg (dismissed):

PR #1111, opened 2023-05-29
fix: issues in rudder event structure

saikumarrs:
## PR Description

- Set default value `null` for `event` and `properties` fields.
- Used the mutated event from the data plane events queue for event callback invocation.

## Notion ticket

https://www.notion.so/rudderstacks/Fix-event-structure-JS-SDK-v3-6aeadd6ee4824420b0f2d1fdc15b45fe?pvs=4

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security

### DIFF
--- a/packages/analytics-js/src/components/eventManager/utilities.ts
+++ b/packages/analytics-js/src/components/eventManager/utilities.ts
@@ -228,6 +228,15 @@ const getEnrichedEvent = (
   }
 
   const processedEvent = mergeDeepRight(rudderEvent, commonEventData) as RudderEvent;
+  // Set the default values for the event properties
+  // matching with v1.1 payload
+  if (processedEvent.event === undefined) {
+    processedEvent.event = null;
+  }
+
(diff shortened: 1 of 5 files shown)


packages/analytics-js/src/components/eventManager/utilities.ts
```diff
@@ -228,6 +228,15 @@ const getEnrichedEvent = (
   }
 
   const processedEvent = mergeDeepRight(rudderEvent, commonEventData) as RudderEvent;
+  // Set the default values for the event properties
+  // matching with v1.1 payload
+  if (processedEvent.event === undefined) {
```
bardisg, line 233: where does this take values from? it is custom ones provided by event method data?

saikumarrs: Yes, both `event` and `properties` are set in the respective factory methods (partial event object) if set by the user.

bardisg (approved):

PR #1135, opened 2023-06-03
fix(analytics-js): use globalThis instead of window to fix issues in SSR

bardisg:
## PR Description

Refactor to ensure the generated npm package will work when imported as dependency to fraeworks that support SSR like NextJS.

- fixes on types to allow reuse and reduce the shared bundles that are generated by modeule federation
- use globalThis instead of window
- ensure no calls to initialize window APIs (like cookie storage) happens before the load method is called
- fixes on html templates
- bumped most dependencies to the latest

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


saikumarrs (approved):

saikumarrs (approved):

packages/analytics-js-plugins/src/utilities/retryQueue/index.ts
```diff
@@ -327,14 +328,12 @@ class Queue extends Emitter {
...
-        id,
-        validKeys: QueueStatuses,
-      },
-      getStorageEngine('localStorage'),
-    );
+    const other = this.storeManager.setStore({
+      id,
+      name: this.name,
+      validKeys: QueueStatuses,
+      type: 'localStorage',
```
saikumarrs, line 335: This should also respect the `storageType` input in the constructor.

packages/analytics-js-plugins/src/utilities/retryQueue/index.ts
```diff
@@ -462,14 +461,12 @@ class Queue extends Emitter {
...
-              id: parts[1],
-              validKeys: QueueStatuses,
-            },
-            getStorageEngine('localStorage'),
-          ),
+          this.storeManager.setStore({
+            id: parts[1],
+            name,
+            validKeys: QueueStatuses,
+            type: 'localStorage',
```
saikumarrs, line 468: Same as above.

packages/analytics-js-plugins/src/utilities/retryQueue/index.ts
bardisg: on it

packages/analytics-js-plugins/src/utilities/retryQueue/index.ts
bardisg: on it

packages/analytics-js-plugins/src/utilities/retryQueue/index.ts
bardisg: seems that this hardcodded value is as expected asreclaim only happens from localstrage

packages/analytics-js-plugins/src/utilities/retryQueue/index.ts
bardisg: seems that this hardcodded value is as expected asreclaim only happens from localstrage

saikumarrs (approved):

PR #1140, opened 2023-06-08
Writing the conveyed information more precisely

ujjawal-4664:
## Using accurate words to convey our message as it is necessary in a Technical document

'Device mode destinations' typically refers to specific devices or screen sizes used for testing and emulating responsiveness in web development.
Instead we should use 'monitoring tools' to address the words like 'Heap' and 'Fullstory'.


bardisg:
Hello @ujjawal-4664 ,

Thank you for your contribution.

Can you please elaborate on what is the issue you attempt to tackle?

In the context of Rudderstack the term "device mode destinations" reffers to destinations supporting web device mode in particular https://www.rudderstack.com/docs/destinations/streaming-destinations/#web-device-mode-destinations

ujjawal-4664:
@bardisg 
I apologize for the confusion . As of my knowledge Heap and FullStory are not specifically referred to as 'device mode destinations'. Heap and FullStory are customer analytics and session replay tools.

so , I think we need to rephrase the sentence a little bit and use 'monitoring tools' instead of 'device mode destionation' in that particular sentence.

bardisg:
Hello @ujjawal-4664,

I think the confusion here is that probably there is no clear understanding what a destination is for RudderStack and the different types of destinations. A destination is a target to send data to.

Please follow through our documentation in order to understand a bit more. 
Here is all our destinations (integrations) list 
https://www.rudderstack.com/integration/?type=Destination/ 
and some more info
https://www.rudderstack.com/docs/destinations/overview/

Each destination can operate in cloud, device or hybrid connection mode.

The term "device mode destinations" is reffering to any of the supported cloud destinations when they operate in device mode connection mode.

Based on that I do not believe in this case any change is required as the term is irrelevant to the nature of the business logic of the destinations themselves.

PR #1155, opened 2023-06-14
fix: `decryptValue()` return early if value is not string

lxcid:
## PR Description

When we deploy `rudder-sdk-js` we been getting a lot of crash report around 

https://github.com/rudderlabs/rudder-sdk-js/blob/66c998ba2785cd33445119f9468d0d3488b34f38/src/utils/storage/storage.js#L65

Basically, the error is:

`e.substring is not a function`

https://github.com/rudderlabs/rudder-sdk-js/blob/66c998ba2785cd33445119f9468d0d3488b34f38/src/utils/storage/storage.js#L62

It look to me that the follow check is not defensive enough. If `value` is not string but truthful. e.g. an empty `{}`, the code will fallthrough and try to access `{}.substring`, which will crash.

My solution is to switch the string check to ignore all non string `value`, while still skipping empty string. the fallen through code will ensure value is string in this case.

It might still fail `parse()` but at least its trap in try catch and default to null.

This is causing failure to load rudder analytics.

## Security

### DIFF
--- a/src/utils/storage/storage.js
+++ b/src/utils/storage/storage.js
@@ -59,7 +59,7 @@ function trim(value) {
  * @param {*} value
  */
 function decryptValue(value) {
-  if (!value || (typeof value === 'string' && trim(value) === '')) {
+  if (!value || typeof value !== 'string' || trim(value) === '') {
     return value;
   }
   if (value.substring(0, defaults.prefix.length) === defaults.prefix) {


MoumitaM:
Hi @lxcid, Can you please sign the [CLA](https://rudderlabs.wufoo.com/forms/rudderlabs-contributor-license-agreement) so we can move forward with your changes?

lxcid:
Done! :)

lxcid:
Would hope you guys can cut a version for this soon as we are collecting quite a bit of these crashes.

lxcid:
Unit tests are passing but failing at sonar scan with following

`Set the SONAR_TOKEN env variable.`

Not too sure how i can solve that.

bardisg:
> Unit tests are passing but failing at sonar scan with following
> 
> `Set the SONAR_TOKEN env variable.`
> 
> Not too sure how i can solve that.

Hello @lxcid ,

Please do not worry about the sonar scan. Since your PR comes fro another forked repo, you do not have access to our github secrets.

This PR is aproved and we will plan it in the next release

lxcid:
i see thanks! would really hope you guys can cut a release as we have been receiving a chunk of errors (6 digits worth) in our error reporting and losing some analytics quality due to some user unable to load.

we are in a dilemma on how should we approach this: wait for release? silent the error? our cut our own release.

lxcid:
Hi again, sorry can i get an eta on this? i don’t mind not being attributed if it mean i can accelerate the deployment of this fix, would really hope to get a release of this fix soon.

this is quite a critical problem for us right now (loss in data quality and eating into our error reporting quota) and i hope to solve it at root cause instead of working around.

note, this cause some client to not able to initialize/load rudder analytics at all. and even after calling reset, it still fails.

bardisg:
Hello, @lxcid ,

Next planned release is next Monday. 

Can you please provide a public facing link with the website/webapp that has the issue so we can assist figuring out the root cause of the issue?

lxcid:
this is deployed on taskade.com. it is not repo able but i’m able to identify the line of code that is throwing error. we recently upgrade from a way older version of rudder analytics.

basically i have tried resetting rudder analytics when i catch this crash and it didn’t work.

IMG_0399

as you can try to read the minified code, it’s where my changes is.


basically, the previous code check for type string and also make sure the string is not empty.

so if you don’t pass in any string (e.g you pass in object or number) it will crash, because subsequently it assume it’s string and try to access `.substring` function.

tbh, you will easily caught this in typescript compiler. so i do think at some point you guys might wanna slowly migrate over to typescript.

lxcid:
i do understand if you wondering how can the value not be string. i wondering too, but if i’m not wrong i upgrade from pre 1.0 version to latest so i think there might be an old code that causing it.

i think for now at least this solution work best as ultimately, it treat these value as invalid and null it. right now it seems like it crashing straight at load() function.

lxcid:
may i ask will we be expecting a release today? this is really a simple but important fix for us and it been stalling for a week.

bardisg:
Hello @lxcid ,

We release in bi-weekly cadence. Release is planned in next Monday.

What I can do you you today, is to release this as beta untill the next production release on 26th.

You would have to change the url of the script in the loading snippet to take in the beta version or add a proxy to point to the beta version url until the 26th. 

Will this solution work for you?

lxcid:
That would be great help! We ok with a beta version. Is it possible u release through npm? I'll point directly to it. :)

bardisg:
Hello @lxcid,

If you use npm package instead of cdn we do not even need a release. 
You can use patch-package direcly in your project (https://www.npmjs.com/package/patch-package).

- Edit the file: ./node-modules/rudder-sdk-js/index.js
- replace 
```
return!e||"string"==typeof e&&""===Rs(e)
```
with
```
return!e||"string"!==typeof e||""===Rs(e)
```
- run npx patch-package rudder-sdk-js
- add patch-package in postinstall npm script

Does this resolve the issue for you?

lxcid:
thank you. this work for me at the moment. we had deployed the fixed

bardisg:
@lxcid ,

Perfect, will merge this one and it will be releashed in next release, then you can remove the patch whenever you upgrade version in your dependencies

PR #1156, opened 2023-06-14
fix: event filtering for empty or non-string event names 

MoumitaM:
## PR Description

Added a fix for the below use case:
When allowlist is selected and track events are triggered without the event name or the name is not a string we shouldn't let the event pass through that destination.

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


bardisg:
@MoumitaM is this something we need to fix in v1.1 too?

packages/analytics-js-plugins/src/nativeDestinationQueue/utilities.ts
```diff
@@ -13,29 +13,34 @@ import { DEFAULT_QUEUE_OPTIONS } from './constants';
...
   }
 
   const { blacklistedEvents, whitelistedEvents, eventFilteringOption } = dest.config;
-  // TODO: might have to make this logic case-sensitive
-  const formattedEventName = eventName.trim().toUpperCase();
+
   switch (eventFilteringOption) {
     // Blacklist is chosen for filtering events
     case 'blacklistedEvents':
+      if (!eventName || typeof eventName !== 'string') {
```
bardisg, line 26: can the user blacklist all track events that do not have a name?

MoumitaM:
> @MoumitaM is this something we need to fix in v1.1 too?

Yes, But this will be a breaking change

packages/analytics-js-plugins/src/nativeDestinationQueue/utilities.ts
```diff
@@ -13,29 +13,34 @@ import { DEFAULT_QUEUE_OPTIONS } from './constants';
...
+        return false;
+      }
       if (Array.isArray(blacklistedEvents)) {
         return blacklistedEvents.some(
-          eventObj => eventObj.eventName.trim().toUpperCase() === formattedEventName,
+          eventObj => eventObj.eventName.trim().toUpperCase() === eventName.trim().toUpperCase(), // TODO: might have to make this logic case-sensitive
```
bardisg, line 31: lets confirm with Ryan if we should make it case-sensitive

packages/analytics-js-plugins/src/nativeDestinationQueue/utilities.ts
```diff
@@ -13,29 +13,34 @@ import { DEFAULT_QUEUE_OPTIONS } from './constants';
 const getNormalizedQueueOptions = (queueOpts: DestinationsQueueOpts): DestinationsQueueOpts =>
   mergeDeepRight(DEFAULT_QUEUE_OPTIONS, queueOpts);
 
-const isEventDenyListed = (eventName: Nullable<string>, dest: Destination) => {
-  if (!eventName || typeof eventName !== 'string') {
+const isEventDenyListed = (eventType: string, eventName: Nullable<string>, dest: Destination) => {
```
bardisg, line 16: lets add unit tests for these new cases

packages/analytics-js-plugins/src/nativeDestinationQueue/utilities.ts
MoumitaM: No

packages/analytics-js-plugins/src/nativeDestinationQueue/utilities.ts
MoumitaM: Already asked, let's wait for his reply. Right now it is case in-sensitive.

packages/analytics-js-plugins/src/nativeDestinationQueue/utilities.ts
saikumarrs, line 16: `eventType` should be of `RudderEventType` type.

packages/analytics-js-plugins/src/nativeDestinationQueue/utilities.ts
saikumarrs, line 26: Let's extract this check into a method.

packages/analytics-js-plugins/src/nativeDestinationQueue/utilities.ts
MoumitaM: Addressed

MoumitaM: Addressed

packages/analytics-js-plugins/src/nativeDestinationQueue/utilities.ts
MoumitaM: Addressed

saikumarrs (approved):

saikumarrs (approved):

bardisg (approved):

PR #1176, opened 2023-06-22
feat: improve adblocker detection

saikumarrs:
## PR Description

- Adblocker detection improved by replacing script load with a HTTP HEAD request

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


packages/analytics-js/src/components/capabilitiesManager/detection/adBlockers.ts
```diff
@@ -1,28 +1,35 @@
...
 
-    // No need to proceed further for Ad-block errors
-    return '';
-  }
+  // Use source config URL here as it is very unlikely to be blocked by adblockers
+  // Only the extra query param should make it vulnerable to adblockers
+  // This will work even if the users proxies it.
+  // The edge case where this doesn't work is when HEAD method is not allowed by the server (user's)
+  const baseUrl = new URL(state.lifecycle.sourceConfigUrl.value as string);
+  const url = `${baseUrl.origin}${baseUrl.pathname}?view=ad`;
```
bardisg, line 14: can you please double check the lists i added in the ticket to see if view=ad exists in both lists or we need to add more params here?

packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -445,6 +446,28 @@ class Analytics implements IAnalytics {
...
+    // TODO: Maybe we should alter the behavior to send the ad-block page event
+    // even if the SDK is still loaded. It'll be pushed into the to be processed queue.
+
+    // Send automatic ad blocked page event if adblockers are detected on the page
+    if (
+      state.capabilities.isAdBlocked.value === true &&
+      payload.category !== 'RudderJS-Initiated'
+    ) {
+      this.page(
+        pageArgumentsToCallOptions(
```
bardisg, line 459: why you use pageArgumentsToCallOptions? just pass expected PageCallOptions object instead

bardisg: do we need to tranck this once per page call or once pr sdk load?

packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -445,6 +446,28 @@ class Analytics implements IAnalytics {
...
       callback: payload.callback,
     });
+
+    // TODO: Maybe we should alter the behavior to send the ad-block page event
+    // even if the SDK is still loaded. It'll be pushed into the to be processed queue.
+
+    // Send automatic ad blocked page event if adblockers are detected on the page
+    if (
+      state.capabilities.isAdBlocked.value === true &&
+      payload.category !== 'RudderJS-Initiated'
```
bardisg, line 456: where is this value defined and whats its purpose, is it just to avoid infinite loop of submits? Also can this be done as constant please?

packages/analytics-js/src/components/capabilitiesManager/detection/adBlockers.ts
saikumarrs: This is a very prominent way to fetch ads. I've verified that this request gets blocked by more than 90% of the ad blockers.

What are the lists you are talking about? I don't see any in the ticket.

packages/analytics-js/src/components/core/Analytics.ts
saikumarrs: 1. Addressed
2. In v1.1, we were doing it for every page call with the assumption that page calls will be made only once.

packages/analytics-js/src/components/core/Analytics.ts
saikumarrs: Yes, this is to avoid an infinite loop.
I have refactored the code to use constants.

packages/analytics-js/src/components/capabilitiesManager/detection/adBlockers.ts
bardisg: added links in the comments section

packages/analytics-js/src/components/core/Analytics.ts
bardisg: lets keep the same behavior then

bardisg (approved):

packages/analytics-js/src/components/capabilitiesManager/detection/adBlockers.ts
saikumarrs: Thanks. I've added more details in the comment section of the card.

PR #1184, opened 2023-06-26
fix(fbpixel): correctly exclude blacklisted PIIs from Rudder payload

pierreavizou:
## PR Description

**The current PII blacklist mechanism for the Facebook Pixel does not work in a number of ways:** 

- Blacklisted PIIs in the config plane are still sent to FB
- PIIs that should be hashed are sent in plain text
- Whitelist mechanism does not work

_As a result, users of this integration could be in breach of local privacy laws and Meta's T&Cs without knowing it._ 

**This PR brings a few fixes and improvements:** 

- Fix blacklist: PIIs blacklisted in config plane are correctly excluded from payloads sent to FB
- Correctly hash PIIs that are configured to be hashed in config plane
- Fix whitelist mechanism
- Feature: a date field can now also be a PII and be both date-parsed and hashed, useful eg. for birthday etc. 

(I have written the code in functional style out of habit, could rewrite it if you would rather enforce a strictly imperative style)

## Security

### DIFF
--- a/src/integrations/FacebookPixel/browser.js
+++ b/src/integrations/FacebookPixel/browser.js
@@ -635,38 +635,42 @@ class FacebookPixel {
     ];
     const whitelistPiiProperties = this.whitelistPiiProperties || [];
     const blacklistPiiProperties = this.blacklistPiiProperties || [];
-    const customPiiProperties = {};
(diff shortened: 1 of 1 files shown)


MoumitaM:
Closing this as the same work with another PR is merged and released.

PR #1197, opened 2023-07-05
feat: remove crypto based encryption for persistent data

saikumarrs:
## PR Description
- For v3 storage encryption, used 'Base64' encoding technique.
- Added storage migrator plugin.
- Improved `toBase64` and `fromBase64` string utility methods logic to support Unicode characters.

## Notion ticket

https://www.notion.so/rudderstacks/d5d15ce4be354e3caae11d4ea9f41477?v=8081ceba15644c85ad7a2cfb918a9c51&p=2524b4e39a8248128422d0431ff0c298&pm=s

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


bardisg:
Hello @saikumarrs,

The aim was to have 2 encryption plugins and one migration one.

The legacy one offers an easy way for existing clients to retain existing data and keep using it if they want to.

The new one is meant for new customers or customers that want to drop existing data.

next step is to create a plugin to migrate form the legacy to the new one that is meant to be used for some months before it can be removed.

Please keep the legacy and create the new one as a separate plugin

saikumarrs:
> Hello @saikumarrs,
> 
> The aim was to have 2 encryption plugins and one migration one.
> 
> The legacy one offers an easy way for existing clients to retain existing data and keep using it if they want to.
> 
> The new one is meant for new customers or customers that want to drop existing data.
> 
> next step is to create a plugin to migrate form the legacy to the new one that is meant to be used for some months before it can be removed.
> 
> Please keep the legacy and create the new one as a separate plugin

Yes, the scope of the task indicated what you mentioned. However, I have a few reasons for taking this decision:

- Users don't care about the underlying encryption technique used. If some user parsed our cookie data independently of our SDK, it'd be a breaking change. The % of those users will be minimal, and we can document it.
- If the old SDK is still used in a sub-domain site, it'd break the v3 SDK when it corrupts the cookies. However, with the v3 SDK in place, the v1/v1.1 SDK on the sub-domain site will break unless we update the v1.1 SDK also to handle this gracefully. Otherwise, we have to force customers to upgrade all sites at once.
- The plugins themselves are quite small, and it felt like an overhead to download two extra plugins on the page to achieve the existing effect. I understand that this defeats the purpose of our plugin architecture, but in favor of the benefits, I had to deviate.

Having said all of this, LMK if you still want me to revert to the original design.

MoumitaM (dismissed):
LGTM

MoumitaM (approved):

bardisg (approved):

PR #1240, opened 2023-07-24
feat: refactor apis

saikumarrs:
## PR Description

Fixes
- Buffered the APIs `setAnonymousId`, `startSession`, and `endSession` as otherwise, the results would be incorrect.
- `getAnonymousId` and `getSessionId` are removed from the loading snippet as there is no point in buffering these requests.
- Trailing slashes removed from data plane URL.
- `OnLoaded` callback invoked with correct arguments.

Enhancements
- Added reporting for invalid write key in RudderAnalytics.
- Error handling introduced for the life cycle.

## Notion ticket

https://www.notion.so/rudderstacks/Dev-testing-JS-v3-after-refactoring-178bae79fec84679ab3af992207882aa?pvs=4

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -312,20 +323,20 @@ class Analytics implements IAnalytics {
...
    * Trigger onLoaded callback if any is provided in config
    */
-  onLoaded() {
+  onInitialized() {
     // Process any preloaded events
     this.processDataInPreloadBuffer();
 
+    // Execute onLoaded callback if provided in load options
+    if (isFunction(state.loadOptions.value.onLoaded)) {
+      (state.loadOptions.value.onLoaded as OnLoadedCallback)((globalThis as any).rudderanalytics);
```
bardisg, line 332: (globalThis as any).rudderanalytics will not provide the correct instance in case more that one analytics instance is available.

What is the issue we attempt to fix and we changes the argument value here?

packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -401,18 +412,18 @@ class Analytics implements IAnalytics {
...
-    if (!isFunction(callback)) {
-      this.logger.error(READY_API_CALLBACK_ERROR(READY_API));
+    if (!state.lifecycle.loaded.value) {
+      state.eventBuffer.toBeProcessedArray.value.push([type, callback]);
       return;
     }
 
-    if (!state.lifecycle.loaded.value) {
-      state.eventBuffer.toBeProcessedArray.value.push([type, callback]);
+    if (!isFunction(callback)) {
```
bardisg, line 420: why we changed order here? we would not want to process the callback if it is not a function

packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -501,7 +512,10 @@ class Analytics implements IAnalytics {
       this.reset();
     }
 
-    this.userSessionManager?.setUserId(payload.userId);
+    // `null` value indicates that previous user ID needs to be retained
+    if (!isNull(payload.userId)) {
```
bardisg, line 516: why we are not able ti set null userId? what happens if user logs out? it should clear user id somehow correct?

packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -501,7 +512,10 @@ class Analytics implements IAnalytics {
       this.reset();
     }
 
-    this.userSessionManager?.setUserId(payload.userId);
+    // `null` value indicates that previous user ID needs to be retained
```
bardisg, line 515: this does not feel correct logic

packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -547,7 +561,11 @@ class Analytics implements IAnalytics {
       return;
     }
 
-    this.userSessionManager?.setGroupId(payload.groupId);
+    // `null` value indicates that previous group ID needs to be retained
```
bardisg, line 564: again, how can you clear it if that is the case?

packages/analytics-js-common/src/utilities/eventMethodOverloads.ts
```diff
@@ -55,80 +55,87 @@ const pageArgumentsToCallOptions = (
...
+    category: category as string,
+    name: name as string,
+    properties: properties as Nullable<ApiObject>,
+    options: options as Nullable<ApiOptions>,
+  };
 
   if (isFunction(callback)) {
-    payload.category = category as string;
-    payload.name = name as string;
-    payload.properties = clone(properties as Nullable<ApiObject>);
```
bardisg, line 63: why we do not clone anymore? we need this to happen as soon as we start accessing consumer provided values

packages/analytics-js-common/src/utilities/eventMethodOverloads.ts
```diff
@@ -55,80 +55,87 @@ const pageArgumentsToCallOptions = (
   options?: Nullable<ApiOptions> | ApiCallback,
   callback?: ApiCallback,
 ): PageCallOptions => {
-  const payload: PageCallOptions = {};
+  const payload: PageCallOptions = {
```
bardisg, line 58: do the changes in this file pass the sanity suite?

packages/analytics-js-common/src/utilities/eventMethodOverloads.ts
bardisg: what is the bug that we are trying to fix and we had to refactor this critical parth of the app?

packages/analytics-js-common/src/utilities/eventMethodOverloads.ts
saikumarrs: Yes.

packages/analytics-js/src/components/core/Analytics.ts
saikumarrs: Similar to the user ID case above.
https://github.com/rudderlabs/rudder-sdk-js/blob/723955d6f408cf97ec7e1a17c4467abcdb2bff10/src/core/analytics.js#L778

packages/analytics-js/src/components/core/Analytics.ts
saikumarrs: You're probably correct. But this is in line with our [v1.1 implementation](https://github.com/rudderlabs/rudder-sdk-js/blob/723955d6f408cf97ec7e1a17c4467abcdb2bff10/src/core/analytics.js#L708).
Setting `null` as the user ID would not clear the user. Instead, users make a `reset` call.

packages/analytics-js/src/components/core/Analytics.ts
saikumarrs: Correct.
The callback is supposed to forward to the analytics instance. Earlier, we were sending `this` which is the Analytics class object.

packages/analytics-js/src/components/core/Analytics.ts
saikumarrs: Same as above.

packages/analytics-js/src/components/core/Analytics.ts
saikumarrs: How do we solve the specific analytics instance problem?

packages/analytics-js/src/components/core/Analytics.ts
bardisg: Can we bring this into Ryan's attention and see if this is something we want to change in v3?

MoumitaM (approved):

bardisg (approved):

PR #1250, opened 2023-07-25
feat(mixpanel): add alias on sdk load

Gauravudia:
## PR Description

Adding alias call on sdk load for mixpanel with a flag `sendAliasOnLoad` 

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


src/core/analytics.js
```diff
@@ -352,7 +352,7 @@ class Analytics {
               self.failedToBeLoadedIntegration.push(intgInstance);
             }
           }
-        }, 100);
+        }, 1);
```
bardisg, line 355: this is too intensive I belive in terms of cpu consumeption imagine 8 destinations that take 300ms to load. how may times will you invoke this method?

src/integrations/Mixpanel/browser.js
```diff
@@ -23,6 +23,7 @@
 /* eslint-disable no-underscore-dangle */
 /* eslint-disable class-methods-use-this */
 import get from 'get-value';
+import isEmpty from 'lodash.isempty';
```
bardisg, line 26: why add a new lodash method? cant we just use !isNotEmpty instead?

.size-limit.js
```diff
@@ -7,7 +7,7 @@ module.exports = [
     name: 'Core - CDN',
     path: 'dist/legacy/rudder-analytics.min.js',
     gzip: true,
-    limit: '37.2 kB',
+    limit: '40.61 kB',
```
bardisg, line 10: this is too much! what have we added?

bardisg (changes requested):

.size-limit.js
Gauravudia: only import was lodash.empty. `npm run check:size` is showing this size limit, Removed the lodash.empty import and reduced the size limit.

src/integrations/Mixpanel/browser.js
Gauravudia: Removed the lodash import

PR #1254, opened 2023-07-27
feat: mixpanel simplified ID merge

Gauravudia:
## PR Description
For feature INT-163
- Removed the support of alias call when Simplified ID merge is selected.
- Removed the fallback value of anonymousId for `identify` calls since `alias` call is now removed. Mixpanel will take care of merging known users (userId) with anonymous users (deviceId) created by mixpanel sdk.

## Notion ticket

[Ticket link](https://www.notion.so/rudderstacks/Mixpanel-Simplified-ID-Merge-Device-Mode-dda27518bdfa4d08a1496b9c4466df11)

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security

### DIFF
--- a/src/integrations/Mixpanel/browser.js
+++ b/src/integrations/Mixpanel/browser.js
@@ -381,7 +381,7 @@ class Mixpanel {
   alias(rudderElement) {
     logger.debug('in Mixpanel alias');
     if (this.identityMergeApi === 'simplified') {
-      logger.error(
+      logger.debug(
         "===Mixpanel: Event type 'alias' is not supported when 'Simplified ID merge' api is selected in webapp===",
       );
       return;
--- a/src/integrations/Mixpanel/browser.js
+++ b/src/integrations/Mixpanel/browser.js
@@ -79,6 +79,7 @@ class Mixpanel {
     this.areTransformationsConnected =
       destinationInfo && destinationInfo.areTransformationsConnected;
     this.destinationId = destinationInfo && destinationInfo.destinationId;
+    this.identityMergeApi = config.identityMergeApi || 'original';
   }
 
   init() {
@@ -186,12 +187,15 @@ class Mixpanel {
     peopleProperties = extendTraits(peopleProperties);
     const superProperties = parseConfigArray(this.superProperties, 'property');
 
-    // eslint-disable-next-line camelcase
-    const user_id = rudderElement.message.userId || rudderElement.message.anonymousId;
+    let userId = rudderElement.message.userId || rudderElement.message.anonymousId;
+    if (this.identityMergeApi === 'simplified') {
+      // calling mixpanel .identify() only for known users
+      userId = rudderElement.message.userId;
+    }
     let traits = formatTraits(rudderElement.message);
     const { email, username } = traits;
     // id
-    if (user_id) window.mixpanel.identify(user_id);
+    if (userId) window.mixpanel.identify(userId);
 
     // name tag
     const nametag = email || username;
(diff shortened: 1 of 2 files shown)


ItsSudip (approved):

src/integrations/Mixpanel/browser.js
```diff
@@ -376,6 +380,13 @@ class Mixpanel {
    */
   alias(rudderElement) {
     logger.debug('in Mixpanel alias');
+    if (this.identityMergeApi === 'simplified') {
+      logger.error(
```
utsabc, line 384: This should be debug log

Gauravudia: debug log added.

utsabc (dismissed):

src/integrations/Mixpanel/browser.js
```diff
@@ -376,6 +380,13 @@ class Mixpanel {
    */
   alias(rudderElement) {
     logger.debug('in Mixpanel alias');
+    if (this.identityMergeApi === 'simplified') {
+      logger.debug(
+        "===Mixpanel: Event type 'alias' is not supported when 'Simplified ID merge' api is selected in webapp===",
```
koladilip, line 385: https://docs.mixpanel.com/docs/tracking/how-tos/identifying-users#how-does-the-simplified-api-differ-from-the-the-original-api This doesn't say it is not supported, it is just saying alias, identifies,  merge behaves same, is my understand correct?

src/integrations/Mixpanel/browser.js
```diff
@@ -79,6 +79,7 @@ class Mixpanel {
     this.areTransformationsConnected =
       destinationInfo && destinationInfo.areTransformationsConnected;
     this.destinationId = destinationInfo && destinationInfo.destinationId;
+    this.identityMergeApi = config.identityMergeApi || 'original';
```
koladilip, line 82: should we assume this? what if users are using original or simplified?

Gauravudia: default value of `identityMergeApi` is set to `original` to keep the existing behaviour. Simplified is an new ID merge released by mixpanel 2-3 months ago. Currently this feature is available in mixpanel project with opt-in basis.

src/integrations/Mixpanel/browser.js
Gauravudia: alias call is nothing but $create_alias event. Which is deprecated and will ignored by mixpanel.
We can improve the log saying that
`"===Mixpanel: Alias call is deprecated in 'Simplified ID Merge'==="`

koladilip: share documentation about deprecation?

Gauravudia: In the above document you shared. This is what mixpanel has mentioned about the alias call
`In the Simplified API the $identify, $merge, and $create_alias events no longer have any special meaning and will be ignored`

koladilip: it is also talking about $identify too right so that is also not supported? may be we are understand this wrongly?

koladilip: for now it is not not clear from docs but based on testing identify is required

utsabc (dismissed):

koladilip (approved):

MoumitaM (approved):

utsabc (approved):

ItsSudip (approved):

PR #1272, opened 2023-08-02
chore: adobe snippets moved to assets directory

yashasvibajpai:
## PR Description

> Moving the cdn deployed analytics.js and heartbeat.js to cdn.

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


bardisg (changes requested):
These assets are deployed to https://s3.console.aws.amazon.com/s3/buckets/rudderlabs-cdn?region=us-east-1&prefix=adobe-analytics-js/&showversions=false 

This is the adobe-analytics-js in cdn.

Lets add them in src/assets/adobe-analytics-js so we can separate them for deployment in case more assets come in the future

You also need to exclude frim sonar scan in the src/assets folder in all scans

yashasvibajpai:
> Apologies @yashasvibajpai can we move assets folder outside src? easier to exclude from all our tooling like that

Sure, should I keep assets in root or inside /public

MoumitaM (approved):

bardisg (approved):

PR #1298, opened 2023-08-11
feat: validating configs before trying to intialize sdk's

anantjain45823:
## PR Description

Our motive is to validate configs before initialisation for all the integrations. 

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


src/core/analytics.js
```diff
@@ -342,6 +354,7 @@ class Analytics {
               // logger.debug(msg);
               this.errorReporting.leaveBreadcrumb(msg);
               intgInstance = new intMod[modName](intg.config, self, intg.destinationInfo);
+              this.validateConfig(intgInstance.getRequiredConfigs(), intg.config, intg.name);
```
saikumarrs, line 357: `getRequiredConfigs` is not an appropriate term. `getRequiredConfigFields` is probably what you meant to use.

saikumarrs, line 357: Not all destinations have implemented the `getRequiredConfigs` method. So, initialization will fail for them right way.
We should first check if that method is defined.

saikumarrs: Moreover, it'll also not work if the user has locked the integrations version.

src/core/analytics.js
```diff
@@ -195,6 +196,17 @@ class Analytics {
...
   }
 
+  validateConfig(requiredConfigsList, configsFromDashboard, intg) {
+    if(!requiredConfigsList){
+      throw Error(`${intg}: required configs not present`);
+    }
+    requiredConfigsList.forEach(config => {
+      const configVal = configsFromDashboard[config];
+      if (!isDefinedAndNotNull(configVal) || (typeof configVal === "string" && configVal.length === 0))
+        throw Error(`${intg}: ${config} is either not present or does not have valid value`);
```
saikumarrs, line 206: The error message needs to be more helpful to the user.

```suggestion
        throw Error(`${intg}: The destination configuration parameter "${config}" is either unavailable or invalid. Please check your destination settings in the dashboard.`);
```

src/core/analytics.js
```diff
@@ -195,6 +196,17 @@ class Analytics {
     );
   }
 
+  validateConfig(requiredConfigsList, configsFromDashboard, intg) {
+    if(!requiredConfigsList){
+      throw Error(`${intg}: required configs not present`);
```
saikumarrs, line 201: Same as the below comment.

src/integrations/Adroll/browser.js
```diff
@@ -17,6 +17,7 @@ class Adroll {
     this.analytics = analytics;
     this.advId = config.advId;
     this.pixId = config.pixId;
+    this.requiredConfigsList = requiredConfigsList;
```
saikumarrs, line 20: Why does it have to be defined in the class? We don't seem to use this.

saikumarrs (changes requested):
What's the need to duplicate the required fields info in the SDKs and also validate again here? For destinations that are created a long time ago?
CC @krishna2020

saikumarrs:
Also, this is not a `chore`. We're adding a new feature to the SDK. So, let's correct the PR title.

src/integrations/Adroll/browser.js
anantjain45823: Removing

PR #1317, opened 2023-08-17
fix(analytics-js): update context page details in every event creation

bardisg:
## PR Description

The NextJs and Gatsby router does not trigger popstate events thus the context page details in state do not get updated (same would happen for reactJs app with react router).

Seems like getting page details in state on each event is inevitable.
Also, campaign and screen details needs to be determined for each event.

## Notion ticket

Resolves SDK-340

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


bardisg:
We could also use polling to check for changes alongside the statepop event lister instead like https://github.com/KoryNunn/on-url-change/blob/master/index.js.

However I am not big fan of intervals as we may get race conditions

MoumitaM (approved):

PR #1335, opened 2023-08-25
feat: add callback for polyfill load

saikumarrs:
## PR Description

- Added a callback function to the default polyfill URL to determine the success of polyfill load.

## Notion ticket

https://linear.app/rudderstack/issue/SDK-365/js-sdk-v3-check-if-polyfills-are-evaluated-properly-with-a-callback

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security

### DIFF
--- a/packages/analytics-js/src/components/utilities/loadOptions.ts
+++ b/packages/analytics-js/src/components/utilities/loadOptions.ts
@@ -130,27 +130,34 @@ const getSourceConfigURL = (
   let searchParams = defSearchParams;
   let pathname = '/sourceConfig/';
   let hash = '';
-  try {
-    const configUrlInstance = new URL(configUrl);
-    if (!(removeTrailingSlashes(configUrlInstance.pathname) as string).endsWith('/sourceConfig')) {
-      configUrlInstance.pathname = `${
-        removeTrailingSlashes(configUrlInstance.pathname) as string
-      }/sourceConfig/`;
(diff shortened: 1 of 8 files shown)


packages/analytics-js/src/components/utilities/loadOptions.ts
```diff
@@ -130,27 +130,34 @@ const getSourceConfigURL = (
...
-      }/sourceConfig/`;
-    }
-    configUrlInstance.pathname = removeDuplicateSlashes(configUrlInstance.pathname);
-
-    defSearchParams.forEach((value, key) => {
-      if (configUrlInstance.searchParams.get(key) === null) {
-        configUrlInstance.searchParams.set(key, value);
+  // Ideally, this check is not required but URL polyfill
+  // doesn't seem to throw errors for empty URLs
+  // TODO: Need to improve this check to find out if the URL is valid or not
```
bardisg, line 135: Do we realy have too? The customers will not only use IE11. If the url is worng it will throw on other borwsers for them. I would not raise the filesize only for this one if there is ot eazy fix

saikumarrs: It's also needed for other use cases as well. So, we have to explore some really light weight solutions.

bardisg (approved):

PR #1348, opened 2023-08-29
fix: customerIo: remove anonymousId as fallback in identify call

anantjain45823:
## PR Description
Resolves INT-491
Removing anonymousId as fallback for identify call in customerIo to maintain parity between cloud and device mode and not create duplicate user

## Notion ticket

[Ticket link
](https://linear.app/rudderstack/issue/INT-491/remove-anonid-as-fallback-from-device-mode)
## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


src/integrations/CustomerIO/browser.js
```diff
@@ -30,10 +30,11 @@ class CustomerIO {
...
   identify(rudderElement) {
     logger.debug('in Customer IO identify');
-    const userId = rudderElement.message.userId
-      ? rudderElement.message.userId
-      : rudderElement.message.anonymousId;
-    const traits = rudderElement.message.context.traits ? rudderElement.message.context.traits : {};
+    const { userId, context } = rudderElement.message;
+    const { traits } = context || {};
+    if(!userId){
+      logger.error('userId is required for Identify call.');
```
sanpj2292, line 36: We don't need to return here ? 🤔

anantjain45823: Yes we do. Added now

sanpj2292: Will userId be only available in `message.userId` or should we consider something others as well (like how we do in [GenericFieldMapping.json](https://github.com/rudderlabs/rudder-transformer/blob/develop/src/v0/util/data/GenericFieldMapping.json#L12-L26))

anantjain45823: In most of the integrations we search for it at root level only but yes w can use GenericFieldMapping as well. Shall I make the change

sanpj2292: It is fine if it is being done for most of the integrations

sanpj2292 (approved):

MoumitaM (approved):

ItsSudip (approved):

PR #1379, opened 2023-09-15
feat: optimize plugin chunks

saikumarrs:
## PR Description

## Notion ticket

Ticket link

## Screenshots

- Chrome
  >
- Firefox
  >
- Safari
  >

## Security


packages/analytics-js-common/src/types/LoadOptions.ts
```diff
@@ -127,4 +144,6 @@ export type LoadOptions = {
   dataPlaneEventsBufferTimeout?: number;
   storage?: StorageOpts;
   preConsent?: PreConsentOptions;
+  // transport mechanism to be used for sending batched requests
+  transportMode?: EventsTransportMode; // Unused for now. This will deprecate the useBeacon and beaconQueueOptions
```
bardisg, line 148: Do we need to add a breaking change for this in the docs?

packages/analytics-js-plugins/package.json
```diff
@@ -73,6 +73,7 @@
     "@preact/signals-core": "1.4.0",
     "@rudderstack/analytics-js-common": "3.0.0-beta.9",
     "crypto-js": "4.1.1",
```
bardisg, line 75: can we now remove this one?

packages/analytics-js-common/src/types/LoadOptions.ts
saikumarrs: This option is not implemented yet in the code. So, no changes required.

packages/analytics-js-plugins/package.json
saikumarrs: This package is still used in the v1.1 package. We can remove it once we upgrade that too later.

bardisg: but this is in the package json of the plugins that is not relevant to v1.1

PR #1411, opened 2023-09-27
feat: alter life cycle for pre consent

saikumarrs:
## PR Description

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-456/update-sdk-load-workflow-for-pre-consent-configuration

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Security


packages/analytics-js/src/services/StoreManager/StoreManager.ts
```diff
@@ -96,7 +126,12 @@ class StoreManager implements IStoreManager {
       const key = sessionKey;
       const storageKey = sessionKey;
       const providedStorageType = entries?.[key]?.type;
-      const storageType = providedStorageType || globalStorageType || DEFAULT_STORAGE_TYPE;
+
+      const preConsentStorageType = getStorageTypeFromPreConsent(state, sessionKey);
+
+      // Storage type precedence order: pre-consent strategy > entry type > global type > default
+      const storageType =
+        preConsentStorageType ?? providedStorageType ?? globalStorageType ?? DEFAULT_STORAGE_TYPE;
```
MoumitaM, line 134: Post consent, there will be another storageType preference before preconsent. Am I correct?

MoumitaM (commented):
Looks good to me.

packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -266,15 +274,10 @@ class Analytics implements IAnalytics {
    * Load configuration
    */
   loadConfig() {
-    if (!state.lifecycle.writeKey.value) {
-      this.errorHandler.onError(
-        new Error('A write key is required to load the SDK. Please provide a valid write key.'),
-        LOAD_CONFIGURATION,
-      );
```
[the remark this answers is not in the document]
saikumarrs: We're already validating the write key in RudderAnalytics. Hence, this is redundant.

packages/analytics-js/src/services/StoreManager/StoreManager.ts
saikumarrs: Yes, it'll take precedence.

MoumitaM (approved):

packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -288,20 +291,14 @@ class Analytics implements IAnalytics {
     this.storeManager?.init();
     this.userSessionManager?.init();
 
-    // Initialize consent manager
+    // Initialize the appropriate consent manager plugin
     if (state.consents.activeConsentManagerPluginName.value) {
-      this.pluginsManager?.invokeSingle(
-        `consentManager.init`,
-        state,
-        this.storeManager,
```
bardisg, line 296: storeManager not required anymore? in the plugins?

packages/analytics-js/src/components/eventManager/utilities.ts
```diff
@@ -247,7 +247,9 @@ const getEnrichedEvent = (
       sessionId: state.session.sessionInfo.value.id || undefined,
       sessionStart: state.session.sessionInfo.value.sessionStart || undefined,
       consentManagement: {
-        deniedConsentIds: clone(state.consents.data.value.deniedConsentIds),
+        deniedConsents: clone(state.consents.data.value.deniedConsents),
+        // For backward compatibility with older rudder-server versions
+        deniedConsentIds: clone(state.consents.data.value.deniedConsents),
```
bardisg, line 252: why we changed that? We explicitly had the suffic "Ids" to clearly state that the values passed by the user should be the ids for the categories/puproses/whatever for their third party consent manager and not the lables/names.

This stands as a constant reminder for developers that these should be the Ids.

It is more appropriate to rename the allowedConcents to allowedConcentIds that changing deniedConsentIds

packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -188,8 +191,13 @@ class Analytics implements IAnalytics {
             this.onInitialized();
             break;
           case 'loaded':
-            this.loadDestinations();
             this.processBufferedEvents();
+            // Short-circuit the life cycle and move to the ready state if pre-consent behavior is enabled
+            if (state.consents.preConsent.value.enabled === true) {
+              state.lifecycle.status.value = 'ready';
+            } else {
+              this.loadDestinations();
```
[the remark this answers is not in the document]
saikumarrs: Yes, because, both the queues would have been initialized and can consume events. While the cloud mode queue might start processing events, the device mode queue is only started after destinations are initialized.

packages/analytics-js/src/components/core/Analytics.ts
saikumarrs: It is not needed for initialization but only for another extension point (updateConsentsInfo) to fetch the user consents.

packages/analytics-js-common/src/utilities/checks.ts
```diff
@@ -3,29 +3,30 @@
  * @param value input value
  * @returns boolean
  */
-const isFunction = (value: any): boolean =>
+// eslint-disable-next-line @typescript-eslint/ban-types
+const isFunction = (value: any): value is Function =>
```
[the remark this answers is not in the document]
saikumarrs: I've added type guards to avoid redundant type casting, which we generally have to do after using these type-checking utility functions in the code.

packages/analytics-js/src/components/core/Analytics.ts
saikumarrs: Thanks for pointing it out anyway. I've fixed another bug.

packages/analytics-js/src/components/eventManager/utilities.ts
saikumarrs: I was unaware of the decision to put the `ids` suffix explicitly, but it felt more like it was meant for OneTrust, as it supports both names and IDs.

It might be confusing when integrating a CMP that doesn't have the concept of IDs per se. 

So, I thought, using a generic name like `consents` across the board (destinations config, UI, SDK event's context, load options, etc.) would help clear any ambiguity.

packages/analytics-js/src/services/StoreManager/StoreManager.ts
```diff
@@ -64,21 +65,50 @@ class StoreManager implements IStoreManager {
...
-  initClientDataStore() {
+  initClientDataStores() {
+    this.initializeStorageState();
+
+    // TODO: fill in extra config values and bring them in from StoreManagerOptions if needed
+    // TODO: should we pass the keys for all in order to validate or leave free as v1.1?
+
+    // Initializing all the enabled store because previous user data might be in different storage
+    // that needs auto migration
+    const storageTypesRequiringInitialization = [MEMORY_STORAGE];
```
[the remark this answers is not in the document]
saikumarrs: We do this because if storage is switched from say localstorage to cookies in the load options, it should not create a new user session.

packages/analytics-js-common/src/utilities/checks.ts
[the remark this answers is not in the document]
bardisg: got it did not know about this one! Thanks

packages/analytics-js/src/components/eventManager/utilities.ts
bardisg: The same issues came up on firday with ketch. I imagine all consent managers will have an Id and a human redable label. We always need the ID

packages/analytics-js/src/services/StoreManager/StoreManager.ts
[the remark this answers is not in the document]
bardisg: But if the storage changes, should it not mean that we are not allowed by user/customer decition to use the others?

packages/analytics-js/src/components/eventManager/utilities.ts
saikumarrs: Ok, in that case, let's use the IDs suffix everywhere. I'll modify the design doc and other things a bit later.

packages/analytics-js/src/services/StoreManager/StoreManager.ts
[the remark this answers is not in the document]
saikumarrs: The user's decision is only for storing the data. Accessing it was never a problem. Moreover, as part of the migration, we also delete data from the other storages to ensure there is only one copy of the persistent data.

packages/analytics-js/src/components/eventManager/utilities.ts
saikumarrs: Done.

bardisg (approved):

PR #1433, opened 2023-10-09
fix: empty anonymous id after reset call

MoumitaM:
## PR Description

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-738/fix-empty-anonymousid-after-reset-call

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Security

### DIFF
--- a/packages/analytics-js/__fixtures__/fixtures.ts
+++ b/packages/analytics-js/__fixtures__/fixtures.ts
@@ -394,7 +394,7 @@ const loadOptionWithInvalidEntry = {
   },
 };
 
-const anonymousIdWithNoStorageOption = {
+const anonymousIdWithNoStorageEntries = {
   userId: {
     type: 'cookieStorage',
     key: userSessionStorageKeys.userId,
@@ -445,5 +445,5 @@ export {
   entriesWithMixStorage,
   loadOptionWithEntry,
   loadOptionWithInvalidEntry,
-  anonymousIdWithNoStorageOption,
+  anonymousIdWithNoStorageEntries,
 };
--- a/packages/analytics-js/__fixtures__/fixtures.ts
+++ b/packages/analytics-js/__fixtures__/fixtures.ts
@@ -394,6 +394,41 @@ const loadOptionWithInvalidEntry = {
   },
 };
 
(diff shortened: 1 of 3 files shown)


packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
```diff
@@ -510,7 +509,11 @@ class UserSessionManager implements IUserSessionManager {
       state.session.authToken.value = defaultUserSessionValues.authToken;
 
       if (resetAnonymousId) {
-        state.session.anonymousId.value = defaultUserSessionValues.anonymousId;
+        if (this.isPersistenceEnabledForStorageEntry('anonymousId')) {
+          this.setAnonymousId();
+        } else {
+          state.session.anonymousId.value = defaultUserSessionValues.anonymousId;
+        }
```
saikumarrs, line 516: Isn't just this enough? If persistence were not enabled for that field, the value in the state would anyway be the default value.

```suggestion
        // This will generate a new anonymous ID
        this.setAnonymousId();
```

packages/analytics-js/__fixtures__/fixtures.ts
```diff
@@ -394,6 +394,41 @@ const loadOptionWithInvalidEntry = {
   },
 };
 
+const anonymousIdWithNoStorageOption = {
```
saikumarrs, line 397: The rest of the entries can be omitted as they'll be defaulted to cookie storage, correct?

saikumarrs (changes requested):

packages/analytics-js/__fixtures__/fixtures.ts
MoumitaM: This is the final storage state, not load option. I am updating the name.

saikumarrs (approved):

bardisg (approved):

bardisg:
@MoumitaM do we need to add a case in sanity suite for this?

MoumitaM:
> @MoumitaM do we need to add a case in sanity suite for this?

Yes, we can do that.

PR #1441, opened 2023-10-11
fix: gracefully handling state property for pages where optimizely is not loaded

shrouti1507:
## PR Description

In our Optimizely Web destination, when a page loads without Optimizely, users will see an error in their console when we try to run one of our `window?.optimizely?.get()` calls.

We are handling this error gracefully, by introducing try-catch block.

resolves INT-818

## Linear task (optional)

[Linear task link](https://linear.app/rudderstack/issue/INT-825/optimizely-web-bug-on-get-function)

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Security


MoumitaM (changes requested):
you need to make changes in registerCurrentlyActiveCampaigns fn also

packages/analytics-js-integrations/src/integrations/Optimizely/browser.js
```diff
@@ -153,20 +162,34 @@ class Optimizely {
...
-        Object.keys(activeCampaigns).forEach(id => {
-          if (referrer) {
-            newActiveCampaign(id, referrer);
-          } else {
-            newActiveCampaign(id);
-          }
-        });
-      } else {
+      try {
+        const state = window?.optimizely?.get('state');
```
MoumitaM, line 166:
```suggestion
let state;
      try {
        state = window?.optimizely?.get('state');
        } catch(){
        state=undefined
        }
```
Do it this way. No need to repeat the below code
```
window.optimizely.push({
            type: 'addListener',
            filter: {
              type: 'lifecycle',
              name: 'initialized',
            },
            handler() {
              checkReferrer();
            },
          });
        }
```

MoumitaM (changes requested):

MoumitaM (approved):

Gauravudia (approved):

packages/analytics-js-integrations/src/integrations/Optimizely/browser.js
```diff
@@ -112,28 +112,37 @@ class Optimizely {
...
     const newActiveCampaign = (id, referrer) => {
-      const state = window?.optimizely?.get('state');
-      if (state) {
-        const activeCampaigns = state.getCampaignStates({
-          isActive: true,
-        });
-        const campaignState = activeCampaigns[id];
-        if (referrer) campaignState.experiment.referrer = referrer;
-        sendCampaignData(campaignState);
+      try {
```
krishna2020, line 115: can we check if reference of optimizely object can be avoided in init flow or whether this code is reachable ? 
The purpose of `initOptimizelyIntegration` itself is to initialise the object & we are trying to access the object within the init flow - are we missing something ?

krishna2020: can do a check `if window.optimizely` instead

shrouti1507: It will provide ambiguous result. Because [here](https://github.com/rudderlabs/rudder-sdk-js/pull/1441) we are already assigning it as empty array. And `if window.optimizely` it will always return true.

krishna2020: ok, we need code refactoring for this destination - many flaws.
If we had `state` in windows.optimizely object, this means it's a json object. Now in this case, when `registerFutureActiveCampaigns` is called, it adding an object which will definitely fail. Check if we need to add try-catch here too

krishna2020 (approved):

MoumitaM (approved):

sandeepdsvs (approved):

bardisg (approved):

PR #1442, opened 2023-10-11
Add explicit export for bundled package

riceboyler:
## PR Description

Adds an explicit export to `analytics-js/package.json` that works properly with NX monorepos (due to them not appearing to properly handle aliased exports). This is entirely additive and should not affect existing versions when using `@rudderstack/analytics-js/bundled` but provides a workaround for people using Next.js with NX.

## Cross Browser Tests

- [x] Chrome
- [x] Firefox
- [ ] IE11 (on a Mac, don't have an easy way to test IE11. And come on, IE11? Really? ;) )

## Security

### DIFF
--- a/packages/analytics-js/package.json
+++ b/packages/analytics-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@rudderstack/analytics-js",
-  "version": "3.0.0-beta.12",
+  "version": "3.0.0-beta.13",
   "description": "RudderStack Javascript SDK",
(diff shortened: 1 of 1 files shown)


bardisg:
I am keeping this one on hold untill we investigate on if we can fix the NX issue with conditional package exports.

Patch package can be used doe anyone affected

bardisg:
With https://github.com/rudderlabs/rudder-sdk-js/pull/1466 you will not need to use /bundled

PR #1462, opened 2023-10-18
refactor(monorepo): verbatimModuleSyntax & no side effect type imports

bardisg:
## PR Description

Use stricter type checks with verbatimModuleSyntax and add linting rules for no side effect type imports
https://johnnyreilly.com/typescript-5-importsnotusedasvalues-error-eslint-consistent-type-imports

## Linear task (optional)

[Linear task link](https://linear.app/rudderstack/issue/SDK-787/improve-typescript-strictness)

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Security


package.json
```diff
@@ -208,7 +208,8 @@
   "overrides": {
     "@babel/traverse": "7.23.2",
     "@nx/devkit": "17.0.2",
-    "debug": "4.3.4"
+    "debug": "4.3.4",
+    "rollup": "4.1.4"
```
saikumarrs, line 212: We're overriding the same version specified in the dependencies. Is this intentional?

saikumarrs (approved):

package.json
bardisg: yes one of the rollup plugins has not caught up yet with v4 in its dependencies (but has no compatiblity issues)

PR #1468, opened 2023-10-25
feat: configure cookie to be fetched from exact domain

MoumitaM:
## PR Description

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Security


saikumarrs (dismissed):

bardisg:
@MoumitaM does this cover both not adding cookies in top domain and not reading values from top domain if the new flag is true?

MoumitaM:
> @MoumitaM does this cover both not adding cookies in top domain and not reading values from top domain if the new flag is true?
With this feature, cookies will only be set from the current exact domain and read from that domain only. Say cookies set in testwebsite.com can only be accessed by testwebsite.com not by any other sub domains. The same is true for sub domains also.

packages/analytics-js-common/src/types/LoadOptions.ts
```diff
@@ -145,6 +145,7 @@ export type LoadOptions = {
   // transport mechanism to be used for sending batched requests
   transportMode?: EventsTransportMode; // Unused for now. This will deprecate the useBeacon and beaconQueueOptions
   consentManagement?: ConsentManagementOptions;
+  sameDomainCookieOnly?: boolean;
```
saikumarrs, line 148:
```suggestion
  sameDomainCookiesOnly?: boolean;
```

saikumarrs (changes requested):

packages/analytics-js-common/src/types/LoadOptions.ts
MoumitaM: Done

saikumarrs (approved):

bardisg (approved):

PR #1476, opened 2023-10-27
feat(analytics-js-integrations): add in-app message in customerIo

ItsSudip:
## PR Description

With these changes, we are adding support for in-app messages to customerIo destination.

## Linear task (optional)

> Resolves INT-52

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


MoumitaM (approved):

yashasvibajpai (commented):
Generic comment:
Do we not customise the layout or design of the in-app message that pops up, like we do in case of Iterable?
Also, how does this in-app message show up?

ItsSudip:
> Do we not customise the layout or design of the in-app message that pops up, like we do in case of Iterable?
Also, how does this in-app message show up?

We can customise the layout on customerio dashboard. If we enable this option then it will automatically show up when we set the in-app message on customerio.

yashasvibajpai (approved):

ujjwal-ab (approved):

PR #1483, opened 2023-10-31
feat(analytics-js-loading-scripts): display info with snippet version

bardisg:
## PR Description

Add version in the generated loaded snippet. Version is also available in the Bugsnug meta and the context lbrary object.

Also updated some dependencies to latest

## Linear task (optional)

[Linear task link](https://linear.app/rudderstack/issue/SDK-761/add-loading-snippet-version-in-snippet)

## Cross Browser Tests

- [x] Chrome
- [x] Firefox
- [x] IE11

## Sanity Suite

## Security


packages/loading-scripts/src/index.ts
```diff
@@ -12,6 +12,8 @@ import type {
   RudderAnalyticsPreloader,
 } from '@rudderstack/analytics-js';
 
+const snippetVersion = '__PACKAGE_VERSION__';
+console.info(`Rudderstack loading snippet: v${snippetVersion}`);
```
saikumarrs, line 16:
```suggestion
console.info(`RudderStack loading snippet: v${snippetVersion}`);
```

packages/loading-scripts/src/index.ts
```diff
@@ -12,6 +12,8 @@ import type {
   RudderAnalyticsPreloader,
 } from '@rudderstack/analytics-js';
 
+const snippetVersion = '__PACKAGE_VERSION__';
```
saikumarrs, line 15: It'd be great if this snippet version were set as a property of the analytics object itself. Might be helpful to support multiple snippet versions etc.
Anyway, it's not super urgent.

saikumarrs (approved):

packages/loading-scripts/src/index.ts
bardisg: why whould we need multible snippet versions?

saikumarrs: Recently, we realized a bug and updated our snippet. It might be possible that newer snippet versions may not be backward compatible with older SDKs versions. I understand that in these situations, the SDK would've also leaped forward in the major version.

However, the SDK knowing the snippet version might help tune itself accordingly if needed. I'm not saying there will be a need like this, but it is also helpful for other things like the snippet version can be included in the reporting data, etc.

bardisg: Hm hm..... would you see this information more usefull in event library object or in bugnag meta?

saikumarrs: Actually, both as in Bugsnag we don't have access to the event data itself.

bardisg: Added in both

saikumarrs (approved):

PR #1502, opened 2023-11-10
fix: preloaded event not containing denied consent ids

MoumitaM:
## PR Description

- Modified existing logic to
     - Set the `loaded` flag to true
     - Execute the preloaded events 
     After denied consent IDs are set in the class variable

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


bardisg (approved):

packages/analytics-v1.1/src/core/analytics.js
```diff
@@ -293,6 +284,18 @@ class Analytics {
...
+
+      // Execute onLoaded callback if provided in load options
+      if (this.options && typeof this.options.onLoaded === 'function') {
+        this.options.onLoaded(this);
+      }
+
+      // Execute any pending buffered requests
+      // (needed if the load call was not previously buffered)
+      processDataInAnalyticsArray(this);
+
```
saikumarrs, line 298: Let's move this block to the end of the function. It's the ideal place as the SDK is now loaded and en route to being ready.

Also, please verify that no other blocks of code before it relies on the `loaded` flag to be `true`.

saikumarrs (changes requested):
Added comments for improvement.

packages/analytics-v1.1/src/core/analytics.js
```diff
@@ -293,6 +284,18 @@ class Analytics {
         }
       }
 
+      // set loaded flag to true
+      this.loaded = true;
+
+      // Execute onLoaded callback if provided in load options
+      if (this.options && typeof this.options.onLoaded === 'function') {
+        this.options.onLoaded(this);
+      }
```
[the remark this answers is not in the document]
MoumitaM: We shouldn't do that. The reason we introduce this `onLoaded` option is to give our customers the flexibility to perform certain things before device mode loading or any event is sent. 
They might be setting a new anonymousId.

packages/analytics-v1.1/src/core/analytics.js
MoumitaM: Which block? line 287-297?

packages/analytics-v1.1/src/core/analytics.js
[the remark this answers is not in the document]
saikumarrs: The device mode integration SDKs load will not finish by the end of this method; we're only starting the load and asynchronously waiting for them.
So, it's not a problem.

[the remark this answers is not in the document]
saikumarrs: As far as the pre-load processing goes, it'd be ideal to flush them before the user can add any new events in the `onLoaded` callback.

Similar to the device mode init, we'd just queue these events to the delivery. So, it won't be a blocking call.

packages/analytics-v1.1/src/core/analytics.js
saikumarrs: Yes.

bardisg: Are we good with this one now?

saikumarrs (approved):

PR #1516, opened 2023-11-16
fix: initialisation of bugsnag in chrome extension

MoumitaM:
## PR Description

Not initialize Bugsnag if the SDK is running on Chrome extension.

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


packages/analytics-js-common/src/utilities/checks.ts
```diff
@@ -64,6 +64,13 @@ const isDefinedNotNullAndNotEmptyString = (value: any): boolean =>
  */
 const isTypeOfError = (obj: any): obj is Error => obj instanceof Error;
 
+/**
+ * Determines if the SDK is running inside a chrome extension
+ * @returns boolean
+ */
+const isSDKRunningInChromeExtension = (): boolean =>
```
bardisg, line 71: This method is more suited for the capabilities detection utils, lets split in a new file please

bardisg: dhecks are for value/type checks

MoumitaM: Done

saikumarrs: I'd make this a state entry like `appEnvironment`. Maybe in the future, we can also collect information like the framework used, etc.

saikumarrs: CC @bardisg

packages/analytics-js-common/__tests__/utilities/detect.test.ts
```diff
@@ -0,0 +1,18 @@
+import { isSDKRunningInChromeExtension } from '../../src/utilities/detect';
+
+describe('Detect env/feature/capabilities', () => {
+  describe('isSDKRunningInChromeExtension', () => {
+    it('should return true if SDK is running inside chrome extension', () => {
+      (window as any).chrome = {
```
saikumarrs, line 6: Let's undo this change after the test is done, please.
Each test case should have a clean and known start.

packages/analytics-js-common/src/utilities/checks.ts
bardisg: Lets see if we need something like that as part of the JS metrics project

packages/analytics-js-common/__tests__/utilities/detect.test.ts
MoumitaM: Done

saikumarrs (approved):

PR #1522, opened 2023-11-22
fix: multiple onReady invocation

MoumitaM:
## PR Description

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

### DIFF
--- a/packages/analytics-js/src/components/core/Analytics.ts
+++ b/packages/analytics-js/src/components/core/Analytics.ts
@@ -175,7 +175,9 @@ class Analytics implements IAnalytics {
             this.onDestinationsReady();
             break;
           case 'ready':
-            this.onReady();
+            (globalThis as typeof window).setTimeout(() => {
+              this.onReady();
+            }, 1);
             break;
           default:
             break;


packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -469,7 +471,7 @@ class Analytics implements IAnalytics {
         this.errorHandler.onError(err, ANALYTICS_CORE, READY_CALLBACK_INVOKE_ERROR);
       }
     } else {
-      state.eventBuffer.readyCallbacksArray.value.push(callback);
+      state.eventBuffer.readyCallbacksArray.value = [callback];
```
bardisg, line 474: you need to create a new array that appends the item I believe. 
Now you restrict that only one on ready callback can be provided by customer

suggestion:

```
state.eventBuffer.readyCallbacksArray.value = [
  ...state.eventBuffer.readyCallbacksArray.value,
  callback
]
```

bardisg (changes requested):

packages/analytics-js/src/components/core/Analytics.ts
MoumitaM: Thank you for pointing it out.

bardisg: I went thorugh all the existing codebase and we do not do any .push in signals anywhere else also

bardisg (dismissed):

bardisg (approved):

saikumarrs (approved):

PR #1523, opened 2023-11-27
fix(analytics-js-integrations): resolved tech debt items

mihir-4116:
## PR Description

- This pull request addresses below changes

1. Removed exceptions thrown to core sdk from integrations
2. Removed reference of analytics object from event level 
3. Resolved eslint errors for integrations

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


packages/analytics-js-integrations/src/integrations/DCMFloodlight/utils.js
```diff
@@ -99,7 +99,7 @@ const mapFlagValue = (key, value) => {
     return 0;
   }
 
-  throw Error(`[DCM Floodlight]:: ${key}: valid parameters are [1|true] or [0|false]`);
+  return 0;
```
shrouti1507, line 102: does this mean we are not offering any visibility to the user about why the event is dropped?

packages/analytics-js-integrations/src/integrations/PinterestTag/browser.js
```diff
@@ -230,8 +230,9 @@ export default class PinterestTag {
     window.pintrk('track', event, pageObject);
   }
 
-  identify() {
-    const userTraits = this.analytics.getUserTraits();
+  identify(rudderElement) {
+    const { context } = rudderElement.message;
+    const userTraits = context?.traits || {};
```
ItsSudip, line 235: Don't we need to consider traits.email?

packages/analytics-js-integrations/src/integrations/DCMFloodlight/utils.js
mihir-4116: we can use logger.error to show error message but we can't throw an exception to core SDK. in this case we can always fall back to default value (which is 0) so there is no need to throw an error

packages/analytics-js-integrations/src/integrations/PinterestTag/browser.js
mihir-4116: no in js-sdk user traits or identify call should be referred from context.traits only and for group call we should refer from root level traits. no need to take it from traits.email

ItsSudip (approved):

shrouti1507 (approved):

koladilip (approved):

PR #1525, opened 2023-11-30
fix: replace stringify with stringifyWithoutCircular

MoumitaM:
## PR Description

- replace JSON.stringify with stringifyWithoutCircular.
- Update logic in stringifyWithoutCircularV1 to match with stringifyWithoutCircular(used in v3)
- update the implementation where stringifyWithoutCircularV1 is used

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

### DIFF
--- a/packages/analytics-js/.size-limit.js
+++ b/packages/analytics-js/.size-limit.js
@@ -31,6 +31,6 @@ module.exports = [
     name: 'Core - CDN',
     path: 'dist/cdn/modern/iife/rsa.min.js',
     gzip: true,
-    limit: '25.5 KiB',
+    limit: '26 KiB',
   },
 ];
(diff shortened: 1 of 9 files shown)


packages/analytics-js/.size-limit.js
```diff
@@ -31,6 +31,6 @@ module.exports = [
     name: 'Core - CDN',
     path: 'dist/cdn/modern/iife/rsa.min.js',
     gzip: true,
-    limit: '25.5 KiB',
+    limit: '26 KiB',
```
saikumarrs, line 34: Why was the size bumped for v3? No related files have been modified.

saikumarrs (approved):
When validating for `null` return value (from the stringify function), we're silently skipping the operation. Instead, we should log proper errors indicating the failed operation.

saikumarrs (approved):

PR #1527, opened 2023-11-30
fix(analytics-js-integrations): remove updating user_properties as part of loadScript call for ga4

mihir-4116:
## PR Description

- This pr addresses removal of the capability to update user_properties within the loadScript call
- In GA4, the persistence of user_properties is ensured once an identify call is made. This persistence spans the entirety of a user's lifetime within GA4
- Henceforth, updates to user_properties will be exclusively managed through the identify call
- Additionally, updating user_properties via loadScript presents cross-domain issues that are better addressed through the identify call

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


utsabc (approved):

MoumitaM (approved):

koladilip (approved):
How do we know that it will not effect existing customers?

PR #1541, opened 2023-12-08
fix: infinite recursion error by tracking visited nodes

koladilip:
## PR Description

This is reported by bugsnag and it is causing max stack size reached error when circular objects are passed.
Fixes INT-1172

## Linear task (optional)

[Linear task link](https://linear.app/rudderstack/issue/INT-1172/fix-infinite-recursion-error-in-js-sdk-flatten-json-function)

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


mihir-4116 (dismissed):

saikumarrs (dismissed):

mihir-4116 (approved):

packages/analytics-js-integrations/__tests__/utils/utils.test.js
```diff
@@ -303,6 +303,7 @@ describe('flattenJsonPayload Tests', () => {
     const result = utils.flattenJsonPayload(testObj);
     expect(result).toStrictEqual({ '': [] });
   });
+
```
ItsSudip, line 306:
```suggestion

```

packages/analytics-js-integrations/src/utils/utils.js
```diff
@@ -126,22 +126,42 @@ function rejectArr(obj, fn) {
...
+    for (let i = 0; i < l; i += 1) {
+      recurse(cur[i], prop ? `${prop}.${i}` : `${i}`, res, visited);
+    }
     if (l === 0) res[prop] = [];
-  } else {
+  };
+
+  const processObject = () => {
     let isEmpty = true;
     Object.keys(cur).forEach(key => {
```
ItsSudip, line 142: Can't we use for...in here?

koladilip: what would the advantage of it, it is just style right?

koladilip: for..in actually showing a warning to use forEach

ItsSudip (approved):

saikumarrs (approved):

PR #1557, opened 2023-12-15
fix(clevertap): region undefined issue

mihir-4116:
## PR Description

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

### DIFF
--- a/packages/analytics-js-integrations/__tests__/integrations/Clevertap/browser.test.js
+++ b/packages/analytics-js-integrations/__tests__/integrations/Clevertap/browser.test.js
@@ -1,0 +1,43 @@
+import Clevertap from '../../../src/integrations/Clevertap/browser';
+
+beforeEach(() => {
+  // Add a dummy script as it is required by the init script
+  const scriptElement = document.createElement('script');
+  scriptElement.type = 'text/javascript';
+  scriptElement.id = 'dummyScript';
+  const headElements = document.getElementsByTagName('head');
+  headElements[0].insertBefore(scriptElement, headElements[0].firstChild);
+});
+
+afterEach(() => {
+  // Reset DOM to original state
+  document.getElementById('dummyScript')?.remove();
+});
+
+afterAll(() => {
+  jest.restoreAllMocks();
+});
+
+describe('Clevertap init tests', () => {
+  let clevertap;
+
+  test('Testing init call of clevertap without region value', () => {
+    clevertap = new Clevertap(
+      { accountId: '12567839', passcode: '123456' },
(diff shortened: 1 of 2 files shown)


packages/analytics-js-integrations/src/integrations/Clevertap/browser.js
```diff
@@ -67,6 +67,7 @@ class Clevertap {
       onUserLogin: [],
       notifications: [],
       privacy: [],
+      region: [],
```
saikumarrs, line 70: Incorrect.
As per their documentation, it's not an array and we have to define a default value.
https://developer.clevertap.com/docs/web-quickstart-guide#add-region-code

saikumarrs (changes requested):

packages/analytics-js-integrations/src/integrations/Clevertap/browser.js
mihir-4116: addressed, kept default region value as none

saikumarrs: We should not define any default value if the region is not explicitly provided by the user as per their docs. Sorry, my bad.

packages/analytics-js-integrations/src/integrations/Clevertap/browser.js
```diff
@@ -67,6 +67,7 @@ class Clevertap {
       onUserLogin: [],
       notifications: [],
       privacy: [],
+      region: 'none',
```
saikumarrs, line 70: No default value should be set.

mihir-4116: addressed

saikumarrs (approved):

yashasvibajpai (approved):

packages/analytics-js-integrations/__tests__/integrations/Clevertap/browser.test.js
```diff
@@ -0,0 +1,43 @@
+import Clevertap from '../../../src/integrations/Clevertap/browser';
```
koladilip, line 1: I saw region in other destinations also, can this be done as generic test?

mihir-4116: i have added it in our notion doc, let's not do common test in this pr as it differs the purpose of this pr. tests are covered for the changes i have made. will add region common test in other pr which i raised

koladilip (approved):

PR #1558, opened 2023-12-18
fix: tiktok remove lowercasing for custom events

anantjain45823:
## PR Description

Resolves INT-1197

For custom events we are lowering their event name but we should retain the message event case as these are case sensitive at tiktok side

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

### DIFF
--- a/packages/analytics-js-integrations/src/integrations/TiktokAds/browser.js
+++ b/packages/analytics-js-integrations/src/integrations/TiktokAds/browser.js
@@ -75,7 +75,7 @@ class TiktokAds {
       logger.error('Event name is required');
       return;
     }
-    event = event.toLowerCase().trim(); // Using optional chaining as we get sometime non-string event as well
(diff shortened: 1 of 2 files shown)


packages/analytics-js-integrations/src/integrations/TiktokAds/browser.js
```diff
@@ -75,7 +75,7 @@ class TiktokAds {
       logger.error('Event name is required');
       return;
     }
-    event = event.toLowerCase().trim();
+    event = event.toLowerCase().trim(); // Using optional chaining as we get sometime non-string event as well
```
ItsSudip, line 78:
```suggestion
    event = event?.toLowerCase().trim(); // Using optional chaining as we get sometime non-string event as well
```

ItsSudip (approved):

PR #1562, opened 2023-12-21
feat: amplitude add support for unset

anantjain45823:
## PR Description

Adding support for [Amplitude unset field ](https://www.docs.developers.amplitude.com/analytics/apis/http-v2-api/#keys-for-the-event-argument:~:text=exceed%2040%20layers.-,user_properties,-Optional.%20Object.%20A)through Rudderstack Integration Object

## Linear task (optional)
Resolves INT-950

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

### DIFF
--- a/packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
+++ b/packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
@@ -6,7 +6,12 @@ import {
 } from '@rudderstack/analytics-js-common/constants/integrations/Amplitude/constants';
 import Logger from '../../utils/logger';
 import { loadNativeSdk } from './nativeSdkLoader';
-import { getTraitsToSetOnce, getTraitsToIncrement, getDestinationOptions } from './utils';
+import {
+  getTraitsToSetOnce,
+  getTraitsToIncrement,
+  getDestinationOptions,
+  getFieldsToUnset,
+} from './utils';
 
 const logger = new Logger(DISPLAY_NAME);
 
@@ -86,12 +91,12 @@ class Amplitude {
     // rudderElement.message.context will always be present as part of identify event payload.
     const { traits } = rudderElement.message.context;
     const { userId, integrations } = rudderElement.message;
-    const amplitudeIntgConfig = getDestinationOptions(integrations);
-    const fieldsToUnset = amplitudeIntgConfig?.fieldsToUnset || undefined;
+    const fieldsToUnset = getFieldsToUnset(integrations);
     const amplitudeIdentify = new window.amplitude.Identify();
     let sendIdentifyCall = false;
-    if (fieldsToUnset && Array.isArray(fieldsToUnset) && fieldsToUnset.length > 0) {
+    if (fieldsToUnset) {
       sendIdentifyCall = true;
+      // AM Docs: https://amplitude.github.io/Amplitude-JavaScript/Identify/#identifyunset
       fieldsToUnset.forEach(fieldToUnset => {
         amplitudeIdentify.unset(fieldToUnset);
       });
--- a/packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
+++ b/packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
@@ -47,7 +47,7 @@ class Amplitude {
     if (this.analytics.loadIntegration) {
       loadNativeSdk(window, document);
     }
-    
+
     const initOptions = {
       attribution: { disabled: this.attribution, trackNewCampaigns: !this.trackNewCampaigns },
       flushQueueSize: this.flushQueueSize,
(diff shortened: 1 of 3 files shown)


packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
```diff
@@ -85,14 +85,23 @@ class Amplitude {
 
     // rudderElement.message.context will always be present as part of identify event payload.
     const { traits } = rudderElement.message.context;
-    const { userId } = rudderElement.message;
-
+    const { userId, integrations } = rudderElement.message;
```
koladilip, line 88: don't have common library to share functionality between cloud and device mode code flows?

packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
```diff
@@ -85,14 +85,23 @@ class Amplitude {
...
+    const amplitudeIntgConfig = getDestinationOptions(integrations);
```
shrouti1507, line 89: Hi Anant, curious to know about what options we can get from here. Also as I see we are not adding any config here, then how these fieldsToUnset is to be populated.

packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
anantjain45823: No and also there is a difference how unset is done in cloud vs js-sdk

packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
shrouti1507: got clarified from @anantjain45823 that it fetched from integrations object

packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
```diff
@@ -85,14 +85,23 @@ class Amplitude {
...
+    const fieldsToUnset = amplitudeIntgConfig?.fieldsToUnset || undefined;
+    const amplitudeIdentify = new window.amplitude.Identify();
+    let sendIdentifyCall = false;
+    if (fieldsToUnset) {
```
koladilip, line 93: is it possible that it can be empty array?

anantjain45823: Yes it is possible

packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
```diff
@@ -85,14 +85,23 @@ class Amplitude {
...
+      sendIdentifyCall = true;
+      fieldsToUnset.forEach(fieldToUnset => {
+        amplitudeIdentify.unset(fieldToUnset);
```
anantjain45823, line 96: @shrouti1507 we unset the properties from here

packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
sanpj2292: Should this be changed to `Array.isArray(fieldsToUnset)` in that case ?

koladilip: should we also check for length > 0?

packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
sanpj2292, line 96: [Q] Can we collect all the fields to unset and call `amplitudeIdentify.unset` only once ? -- Just asking

packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
anantjain45823: Done

koladilip (approved):

packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
sanpj2292: Don't think so many conditions are required. Just `Array.isArray()` is actually sufficient. But it's fine

sanpj2292 (approved):

packages/analytics-js-integrations/src/integrations/Amplitude/browser.js
koladilip: What about length?

sanpj2292: If type is array
```javascript
> a = []
[]
> a.forEach(console.log)
undefined
```
Nothing would happen

koladilip: what about `sendIdentifyCall = true;`

sanpj2292: Acha.. Got it 👍🏼

koladilip (approved):

PR #1568, opened 2023-12-29
fix: enhance error handling

MoumitaM:
## PR Description

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

### DIFF
--- a/packages/analytics-v1.1/src/core/analytics.js
+++ b/packages/analytics-v1.1/src/core/analytics.js
@@ -607,34 +607,30 @@ class Analytics {
 
     // send the queued events to the fetched integration
     object.toBeProcessedByIntegrationArray.forEach(event => {
-      try {
-        const methodName = event[0];
-        event.shift();
+      const methodName = event[0];
+      event.shift();
 
-        // convert common names to sdk identified name
-        if (Object.keys(event[0].message.integrations).length > 0) {
-          transformToRudderNames(event[0].message.integrations);
-        }
-
-        // if not specified at event level, All: true is default
-        const clientSuppliedIntegrations = event[0].message.integrations;
-
-        // get intersection between config plane native enabled destinations
-        // (which were able to successfully load on the page) vs user supplied integrations
-        const successfulLoadedIntersectClientSuppliedIntegrations = findAllEnabledDestinations(
-          clientSuppliedIntegrations,
-          object.clientIntegrationObjects,
-        );
-
-        // send to all integrations now from the 'toBeProcessedByIntegrationArray' replay queue
-        this.processAndSendEventsToDeviceMode(
-          successfulLoadedIntersectClientSuppliedIntegrations,
-          event[0],
-          methodName,
-        );
-      } catch (e) {
-        handleError(e);
+      // convert common names to sdk identified name
+      if (Object.keys(event[0].message.integrations).length > 0) {
+        transformToRudderNames(event[0].message.integrations);
       }
+
+      // if not specified at event level, All: true is default
+      const clientSuppliedIntegrations = event[0].message.integrations;
+
+      // get intersection between config plane native enabled destinations
+      // (which were able to successfully load on the page) vs user supplied integrations
+      const successfulLoadedIntersectClientSuppliedIntegrations = findAllEnabledDestinations(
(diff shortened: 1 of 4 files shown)


packages/analytics-v1.1/src/core/analytics.js
```diff
@@ -603,30 +607,34 @@ class Analytics {
 
     // send the queued events to the fetched integration
     object.toBeProcessedByIntegrationArray.forEach(event => {
-      const methodName = event[0];
-      event.shift();
+      try {
```
saikumarrs, line 610: This is not necessary. We have to mainly target those code blocks where data is processed for destinations iteratively and those that can potentially prevent the events from going to the data plane.

packages/analytics-v1.1/src/core/analytics.js
```diff
@@ -988,19 +996,23 @@ class Analytics {
...
+            successfulLoadedIntersectClientSuppliedIntegrations,
+            rudderElement,
+            type,
+          );
+        } catch (err) {
+          handleError(err, `processAndSendEventsToDeviceMode::`);
+        }
       }
 
       const clonedRudderElement = R.clone(rudderElement);
```
saikumarrs, line 1018: The code blocks from Ln 994-1016 should be completely isolated with it's own try-catch block around it to prevent any damage to the cloud mode events delivery below.

packages/analytics-v1.1/src/core/analytics.js
```diff
@@ -603,30 +607,34 @@ class Analytics {
 
     // send the queued events to the fetched integration
     object.toBeProcessedByIntegrationArray.forEach(event => {
```
saikumarrs, line 609: Move Ln 606 to the beginning of this function so that delivery to the data plane is unaffected by the destinations.

packages/analytics-v1.1/src/core/analytics.js
saikumarrs: Let's audit the v3 code also to ensure they're isolated.

saikumarrs (changes requested):

packages/analytics-v1.1/src/core/analytics.js
MoumitaM: Aren't CM events dependent on the integration object construction? which is happening on line no 597

packages/analytics-v1.1/src/core/analytics.js
MoumitaM: Addressed

packages/analytics-v1.1/src/core/analytics.js
MoumitaM: Addressed

packages/analytics-v1.1/src/core/analytics.js
saikumarrs: You are probably right. So you can isolate that above block with error handling.

packages/analytics-v1.1/src/core/analytics.js
saikumarrs: We can add error handling in the event repository (v3) around the individual plugin invocations.

MoumitaM: Addressed

packages/analytics-v1.1/src/core/analytics.js
MoumitaM: Addressed

packages/analytics-v1.1/src/core/analytics.js
```diff
@@ -587,19 +591,23 @@ class Analytics {
...
+          this.integrationsData,
+          object.clientIntegrationObjects,
+        );
+        object.clientIntegrationsReady = true;
+        // Execute the callbacks if any
+        object.executeReadyCallback();
+      }
 
-    this.processBufferedCloudModeEvents();
+      this.processBufferedCloudModeEvents();
```
saikumarrs, line 607: Let's keep this outside of this try-catch. Failure to ensure ready status should not hamper the data plane events delivery.

saikumarrs (approved):

PR #1570, opened 2024-01-08
fix(googleads): added validation to discard event if event name is no event name is present

mihir-4116:
## PR Description
- This pr removes unnecessary logger error and checks if event name is present or not.if event name is not present then it will discard google ads event

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


koladilip (commented):
what about test cases?

mihir-4116:
> what about test cases?

done, added test for no event name is present

koladilip (approved):

packages/analytics-js-integrations/__tests__/integrations/GoogleAds/browser.test.js
```diff
@@ -76,6 +98,24 @@ describe('Scenario to test dynamic remarketing event by enabling the dynamicRema
...
 
 // New Config Test Cases
+describe('Edge scenarios', () => {
+  let googleAds;
+  beforeEach(() => {
+    googleAds = new GoogleAds(googleAdsConfigs[2], {  }, destinationInfo);
+    googleAds.init();
+    window.gtag = jest.fn();
+  });
+  test('Scenario to test no event name', () => {
```
ItsSudip, line 108: Can you add a test with page call also? We are using same function in page calls as well.

mihir-4116: addressed

ItsSudip (approved):

PR #1579, opened 2024-01-09
fix: ga4 page call mappings

mihir-4116:
## PR Description

- In GA4 page call if `extendPageViewParams` toggle is disabled then we used to send three properties to GA4 which are page, url and referrer. but there is a typo in mapping and as a result parameters are not going to GA4. instead of properties.url we kept mapping as url only and as a result it never got picked up from payload. we need to fix that

- we use `getValueFromMessage(message, key)`  function to get key value from payload and key place needs to be specified properly. if url is present inside properties then we need to give it as `properties.url`. keeping just `url` will not work

## Linear task (optional)
https://linear.app/rudderstack/issue/INT-1328/ga4-fix-device-mode-page-call-mappings

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


packages/analytics-js-integrations/__tests__/integrations/GA4/__mocks__/data.js
```diff
@@ -1377,7 +1377,7 @@ const trackEvents = [
 
 const pageEvents = [
   {
-    description: 'Call with properties',
+    description: 'Scenario to test page_view event with extendPageViewParams toggle disabled',
```
koladilip, line 1380: you can omit "Scenario to test"

mihir-4116: addressed

utsabc:
Can you explain the tests with GH comments?

mihir-4116:
> Can you explain the tests with GH comments?

We have one config option in webapp for page calls which is `extendPageViewParams`. if that toggle is enabled then in page call we send all properties we receive in payload. if that toggle is disabled then we only send three properties which are `page_location`, `page_referrer` and `page_title`. These 2 test case are ensuring that respective payloads are being sent to GA4 based on toggle values

utsabc (dismissed):

packages/analytics-js-integrations/src/integrations/GA4/config.js
```diff
@@ -120,9 +120,9 @@ const eventParams = {
     sourceKeys: ['properties.item_id', 'properties.cart_id'],
     destKey: 'item_id',
   },
-  url: { sourceKeys: 'url', destKey: 'page_location' },
-  referrer: { sourceKeys: 'referrer', destKey: 'page_referrer' },
-  title: { sourceKeys: 'title', destKey: 'page_title' },
+  url: { sourceKeys: ['properties.url', 'context.page.url'], destKey: 'page_location' },
+  referrer: { sourceKeys: ['properties.referrer', 'context.page.referrer'], destKey: 'page_referrer' },
+  title: { sourceKeys: ['properties.tile', 'context.page.title'], destKey: 'page_title' },
```
saikumarrs, line 125: `tile` is probably a typo.

packages/analytics-js-integrations/__tests__/integrations/GA4/browser.test.js
```diff
@@ -75,8 +80,8 @@ describe('Google Analytics 4 Page events tests', () => {
         expect(error.message).toEqual(output.message);
       }
     });
-  });
-})
+  })
```
saikumarrs, line 83:
```suggestion
  });
```

packages/analytics-js-integrations/__tests__/integrations/GA4/browser.test.js
```diff
@@ -52,18 +52,23 @@ describe('Google Analytics 4 Track events tests', () => {
...
-    ga4.init();
-    window.gtag = jest.fn();
-  });
+  pageEvents.forEach((event, index) => {
+    let ga4;
+    beforeEach(() => {
+      ga4 = new GA4(
+        {
+          measurementId: 'G-123456',
+          extendPageViewParams: index !== 0,
```
saikumarrs, line 61: This will make things hard to maintain. It is better to take that from the input.

packages/analytics-js-integrations/__tests__/integrations/GA4/browser.test.js
```diff
@@ -52,18 +52,23 @@ describe('Google Analytics 4 Track events tests', () => {
...
-  beforeEach(() => {
-    ga4 = new GA4(
-      { measurementId: 'G-123456', extendPageViewParams: true, capturePageView: 'rs', piiPropertiesToIgnore: [{ piiProperty: '' }] },
-      { getUserId: () => '1234', getUserTraits: () => { } },
-      destinationInfo,
-    );
-    ga4.init();
-    window.gtag = jest.fn();
-  });
+  pageEvents.forEach((event, index) => {
```
saikumarrs, line 55: You can use `it.each` for automatic enumeration of test cases.
[optional]

saikumarrs (changes requested):

packages/analytics-js-integrations/__tests__/integrations/GA4/browser.test.js
mihir-4116: addressed

packages/analytics-js-integrations/__tests__/integrations/GA4/browser.test.js
mihir-4116: addressed

packages/analytics-js-integrations/__tests__/integrations/GA4/browser.test.js
mihir-4116: will pick up later and do it for all ga4 tests

saikumarrs (approved):

utsabc (approved):

packages/analytics-js-integrations/__tests__/integrations/GA4/browser.test.js
```diff
@@ -80,15 +71,20 @@ describe('Google Analytics 4 Page events tests', () => {
...
+  });
 });
 
 describe('Google Analytics 4 Group events tests', () => {
   let ga4;
   beforeEach(() => {
     ga4 = new GA4(
-      { measurementId: 'G-123456', extendPageViewParams: true, capturePageView: 'rs', piiPropertiesToIgnore: [{ piiProperty: '' }] },
-      { getUserId: () => '1234', getUserTraits: () => { } },
+      {
```
koladilip, line 81: input.config can be used right?

packages/analytics-js-integrations/__tests__/integrations/GA4/browser.test.js
```diff
@@ -108,15 +104,20 @@ describe('Google Analytics 4 Group events tests', () => {
...
+});
 
 describe('Google Analytics 4 Identify events tests with no piiPropertiesToIgnore', () => {
   let ga4;
   // Config 1 : default piiPropertiesToIgnore value
   beforeEach(() => {
     ga4 = new GA4(
-      { measurementId: 'G-123456', extendPageViewParams: true, capturePageView: 'rs', piiPropertiesToIgnore: [{ piiProperty: '' }] },
-      { getUserId: () => '1234', getUserTraits: () => { } },
+      {
```
koladilip, line 114: input.config can be used?

packages/analytics-js-integrations/__tests__/integrations/GA4/browser.test.js
mihir-4116: we are taking configs from test case because for each test config is changing in page call, for for rest of the call config is same so no need to add it in tests payload

packages/analytics-js-integrations/__tests__/integrations/GA4/browser.test.js
mihir-4116: we are taking configs from test case because for each test config is changing in page call, for for rest of the call config is same so no need to add it in tests payload

koladilip (approved):

PR #1591, opened 2024-01-19
chore(release): pulling v3-release/3.0.0-beta.24 into main

devops-github-rudderstack:
:crown: *An automated PR*


packages/analytics-js-integrations/package.json
```diff
@@ -105,6 +105,7 @@
     "@rudderstack/analytics-js-common": "*",
     "component-each": "0.2.6",
     "crypto-js": "4.2.0",
+    "each": "^2.6.0",
```
bardisg, line 108: what is this new library? Do we need it? 
does https://ramdajs.com/docs/#forEach cover our needs?

bardisg: if we do we should not allow ^ range and it should also be in root level package json

bardisg (approved):

PR #1598, opened 2024-01-28
fix: appcues to support flattening obj/array for identify

yashasvibajpai:
## PR Description
Appcues does not support sending objects to the Appcues.identify() method but we do not provide a method to stringify the data when we send it to the destination. We are flattening the array or object type traits here.

## Linear task (optional)

Resolves INT-1428

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


packages/analytics-js-integrations/__tests__/integrations/Appcues/appcues.test.js
```diff
@@ -0,0 +1,163 @@
...
+
+const mockAppcuesSDK = () => {
+  window.Appcues = {
+    identify: jest.fn(),
+    track: jest.fn(),
+    page: jest.fn(),
+  };
+};
+
+beforeEach(() => {
```
sanpj2292, line 26: Is there any reason we have to do this in `beforeEach` rather than `beforeAll` ?

packages/analytics-js-integrations/__tests__/integrations/Appcues/appcues.test.js
```diff
@@ -0,0 +1,163 @@
...
+  });
+
+  it('Testing init call of Appcues', () => {
+    appcues = new Appcues(appcuesConfig, { loglevel: 'debug' });
+    appcues.init();
+    expect(typeof appcues).toBe('object');
+  });
+});
+
+describe('Amplitude identify tests', () => {
```
sanpj2292, line 50: Amplitude ?

yashasvibajpai: sorry fixed it

packages/analytics-js-integrations/__tests__/integrations/Appcues/appcues.test.js
yashasvibajpai: Yes it was not getting initialised, checked while debugging so did it this way

sanpj2292 (approved):

packages/analytics-js-integrations/src/integrations/Appcues/browser.js
```diff
@@ -54,6 +54,15 @@ class Appcues {
   identify(rudderElement) {
     const { traits } = rudderElement.message.context;
     const { userId } = rudderElement.message;
+    // iterate through traits and flatten any properties that are objects or arrays
+    Object.keys(traits).forEach(key => {
+      if (typeof traits[key] === 'object') {
```
MoumitaM, line 59: typeof `null` also returns "object". modify the if condition.

MoumitaM (changes requested):

packages/analytics-js-integrations/src/integrations/Appcues/browser.js
```diff
@@ -54,6 +55,15 @@ class Appcues {
   identify(rudderElement) {
     const { traits } = rudderElement.message.context;
     const { userId } = rudderElement.message;
+    // iterate through traits and flatten any properties that are objects or arrays
+    Object.keys(traits).forEach(key => {
+      if (typeof traits[key] === 'object' && typeof traits[key] !== null) {
```
MoumitaM, line 60:
```suggestion
      if ( traits[key] && typeof traits[key] === 'object' ) {
```

MoumitaM (changes requested):

MoumitaM (dismissed):

sanpj2292 (approved):

sanpj2292 (approved):

ItsSudip (approved):

PR #1602, opened 2024-01-30
feat(analytics-js): change default value of storage migrate flag to true

bardisg:
## PR Description

Change the value of storage migrate loadOption to true to assist existing users migrate

## Linear task (optional)

[Linear task link](https://linear.app/rudderstack/issue/SDK-1208/set-storageencryptionmigrate-default-value-true)

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


saikumarrs (approved):
Should we also go one step ahead and automatically include the migration plugin if the user provides the `plugins` input and the flag value is true?

PR #1610, opened 2024-02-08
feat: onboard new destination commandbar

anantjain45823:
## PR Description
Onboard new destination commandbar
CommandBar is device mode only destination supports two calls:
trackevent -> for event tracking -> mapped to rudderstack's track call
boot -> for user identification -> mapped to rudderstack's identify call

## Linear task (optional)

Linear task link
Resolves INT-1311

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

### DIFF
--- a/packages/analytics-js-integrations/__tests__/integrations/CommandBar/browser.test.js
+++ b/packages/analytics-js-integrations/__tests__/integrations/CommandBar/browser.test.js
@@ -1,0 +1,115 @@
+import CommandBar from '../../../src/integrations/CommandBar/browser';
+
+afterAll(() => {
+  jest.restoreAllMocks();
+});
+const destinationInfo = {
+  areTransformationsConnected: false,
+  destinationId: 'sample-destination-id',
+};
+
+describe('CommandBar init tests', () => {
+  let Commandbar;
+
+  test('Testing init call of CommandBar', () => {
+    Commandbar = new CommandBar({ orgId: '12567839' }, { loglevel: 'debug' }, destinationInfo);
+    Commandbar.init();
+    expect(typeof window.CommandBar).toBe('object');
+  });
+});
+
+describe('CommandBar Track event', () => {
+  let Commandbar;
+  Commandbar = new CommandBar({ orgId: '12567839' }, { loglevel: 'debug' }, destinationInfo);
+  Commandbar.init();
+  window.CommandBar.trackEvent = jest.fn();
+  test('Testing Track Events', () => {
+    Commandbar.track({
+      message: {
+        context: {},
+        event: 'dummyEventName',
+        properties: {
+          customProp: 'testProp',
+        },
+      },
+    });
+    expect(window.CommandBar.trackEvent.mock.calls[0][0]).toEqual('dummyEventName');
+    expect(window.CommandBar.trackEvent.mock.calls[0][1]).toEqual({});
+  });
+});
+describe('CommandBar Identify event', () => {
+  let Commandbar;
+  Commandbar = new CommandBar({ orgId: '12567839' }, { loglevel: 'debug' }, destinationInfo);
+  Commandbar.init();
+  window.CommandBar.boot = jest.fn();
+  test('Testing Identify sending traits in context and hmacId as well', () => {
+    Commandbar.identify({
+      message: {
+        userId: 'rudder01',
+        context: {
+          traits: {
+            email: 'abc@ruddertack.com',
+            isRudderEvents: true,
+            hmacId: 'hmacUserId',
+          },
+        },
+      },
+    });
+    expect(window.CommandBar.boot.mock.calls[0][0]).toEqual('rudder01');
+    expect(window.CommandBar.boot.mock.calls[0][1]).toEqual({
+      email: 'abc@ruddertack.com',
+      isRudderEvents: true,
+    });
+    expect(window.CommandBar.boot.mock.calls[0][2]).toEqual({
+      hmac: 'hmacUserId',
+    });
+  });
+  test('Testing Identify sending traits in message and hmacId in traits.context ', () => {
+    Commandbar.identify({
+      message: {
+        userId: 'rudder01',
+        context: {
+          traits: {
+            email: 'abc@ruddertack.com',
+            isRudderEvents: true,
+          },
+        },
+        traits: {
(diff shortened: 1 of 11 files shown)


anantjain45823:
I have -n flag commit as commit was failing in local and spending some hours even it did not get resolved and need to check of its a local machine issue

packages/analytics-js-integrations/src/integrations/CommandBar/browser.js
```diff
@@ -0,0 +1,78 @@
...
+      destinationId: this.destinationId,
+    } = destinationInfo ?? {});
+  }
+
+  init() {
+    loadNativeSdk(this.orgId);
+  }
+
+  isLoaded() {
+    return !!(window.CommandBar && typeof window.CommandBar === 'object');
```
saikumarrs, line 31:
```suggestion
    return window.CommandBar && typeof window.CommandBar === 'object';
```

saikumarrs: Moreoever, the checking condition is not correct. `window.CommandBar` is always an object.
There must be something else we can use to check.

saikumarrs (changes requested):

packages/analytics-js-integrations/src/integrations/CommandBar/browser.js
anantjain45823: I checked commandbar documentation to check for some other function like isReady or isLoaded but nothing

saikumarrs: Try to check the contents of `window.CommandBar` before and after the load and check if it reveals anything.

Otherwise, let's check with their team.

anantjain45823: Ok

packages/analytics-js-integrations/src/integrations/CommandBar/browser.js
```diff
@@ -0,0 +1,78 @@
...
+      return;
+    }
+
+    const instanceAttributes = {}; // stores commandBar options associated with the current session.
+
+    if (hmacId) {
+      instanceAttributes.hmac = hmacId;
+      delete userTraits.hmacId;
+    }
+    window.CommandBar.boot(userId, userTraits, instanceAttributes);
```
saikumarrs, line 60: Should we call `boot` twice if a user is identified twice but the second time with only additional traits?

I believe booting should happen only once. Correct me if I'm wrong.

anantjain45823: Yes booting should happen only once.
But can't that be handled at implementation level.
Also how will we knoe if a particular user is being identified or not. Shall we use cookie for this ?

packages/analytics-js-integrations/src/integrations/CommandBar/browser.js
anantjain45823: This is done

packages/analytics-js-integrations/src/integrations/CommandBar/browser.js
saikumarrs: Users can make mistakes. However, let's check with the CB team what happens when we boot with the same user ID but a different (extra) set of traits.

If this is a problem, then we can store if a user was identified and handle it accordingly.

anantjain45823: Starting a thread with CB team

anantjain45823: Confirmed with CB team it is update or create function hence we can call multiple times

packages/analytics-js-integrations/src/integrations/CommandBar/browser.js
```diff
@@ -0,0 +1,78 @@
...
+   * @returns
+   */
+  identify(rudderElement) {
+    const { message } = rudderElement;
+    const { traits, userId, context } = message;
+    const userTraits = { ...context.traits, ...traits };
+    const { hmacId } = userTraits;
+    if (!userId) {
+      logger.error('userId from identify call is missing');
+      return;
```
shrouti1507, line 51: will this actually drop the event? @MoumitaM to confirm

packages/analytics-js-integrations/__tests__/integrations/CommandBar/browser.test.js
```diff
@@ -0,0 +1,115 @@
...
+    expect(window.CommandBar.boot.mock.calls[0][0]).toEqual('rudder01');
+    expect(window.CommandBar.boot.mock.calls[0][1]).toEqual({
+      email: 'abc@ruddertack.com',
+      isRudderEvents: true,
+    });
+    expect(window.CommandBar.boot.mock.calls[0][2]).toEqual({
+      hmac: 'hmacUserId',
+    });
+  });
+  test('Testing Identify no traits or hmacId passed ', () => {
```
shrouti1507, line 91: Can you elastrate this a bit more? I see `hmacId: 'hmacUserId',` is present in traits and context.traits is also present

packages/analytics-js-integrations/src/integrations/CommandBar/browser.js
anantjain45823: It wont actually drop any event but yeah will not make any call to commandbar as before reaching that statement it will return

packages/analytics-js-integrations/__tests__/integrations/CommandBar/browser.test.js
anantjain45823: Basically fetching traits from both context.traits and traits thats why hmacId can be present anywhere

saikumarrs (approved):

shrouti1507 (approved):

packages/analytics-js-integrations/src/integrations/CommandBar/browser.js
saikumarrs: Yes, it'll drop the event to CommandBar (CB) as we're not reaching the stage where data is forwarded to the CB SDK.

PR #1611, opened 2024-02-11
feat: add ecom spec support to google ads via config

yashasvibajpai:
## PR Description

We are adding support for a v2 toggle from UI, which will allow us to utilise GA4 event specification for all the ecommerce events in Google Ads

## Linear task (optional)
Resolves INT-1573

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


packages/analytics-js-integrations/src/integrations/GoogleAds/browser.js
```diff
@@ -18,6 +18,7 @@ import {
 } from './utils';
 import { loadNativeSdk } from './nativeSdkLoader';
 
+import { prepareParamsAndEventName } from '../GA4/utils';
```
saikumarrs, line 21: If we see this as a helpful utility across Google integrations, we should move it to a different place.
Let's create a Linear ticket if needed. Resolve this comment accordingly.

saikumarrs (dismissed):

ItsSudip (dismissed):

packages/analytics-js-integrations/src/integrations/GoogleAds/browser.js
yashasvibajpai: Sure, will check on this CC @mihir-4116

packages/analytics-js-integrations/src/integrations/GoogleAds/browser.js
```diff
@@ -45,6 +46,7 @@ class GoogleAds {
     // Depreciating: Added to make changes backward compatible
     this.dynamicRemarketing = config.dynamicRemarketing;
     this.allowEnhancedConversions = config.allowEnhancedConversions || false;
+    this.v2 = config.v2 || true;
```
Gauravudia, line 49: Default should be `false` right? To backward compatibility

Gauravudia:
@yashasvibajpai Sonar report some new issues.

packages/analytics-js-integrations/src/integrations/GoogleAds/browser.js
yashasvibajpai: No, for new destinations created going ahead we want to keep it enabled, so that ecom spec is followed.

For existing destinations to avoid any breaking change, we will have a migration to disable it

ItsSudip (approved):

Gauravudia (approved):

PR #1621, opened 2024-02-20
feat: klaviyo profile mapping

Gauravudia:
## PR Description

Resolves INT-1552
Added new profile mappings for klaviyo

postalCode -> `zip`
state -> `region`
street -> `address1`
https://developers.klaviyo.com/en/v2023-02-22/reference/create_profile

## Linear task (optional)

https://linear.app/rudderstack/issue/INT-1552/klaviyo-identify-mappings-for-postalcode-state-street-need-to-be

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


koladilip (approved):

packages/analytics-js-integrations/src/integrations/Klaviyo/browser.js
```diff
@@ -119,8 +120,9 @@ class Klaviyo {
       $country: country,
       $organization: get(message, 'context.traits.organization'),
       $title: get(message, 'context.traits.title'),
-      $region: get(message, 'context.traits.region'),
-      $zip: get(message, 'context.traits.zip'),
+      $region: get(message, 'context.traits.region') || state,
```
shrouti1507, line 123: let's a add small test case where the newly added code is covered

Gauravudia: Klaviyo don't have any testcases at all.

anantjain45823: @Gauravudia It would be better if we can add the test cases in that cases as this is what we are starting doing to cover for integration with no test cases

anantjain45823 (approved):

packages/analytics-js-integrations/src/integrations/Klaviyo/browser.js
Gauravudia: Will raise a separate PR covering this.

shrouti1507 (approved):

PR #1622, opened 2024-02-20
feat: add support of custom page event name in mixpanel

Gauravudia:
## PR Description

Resolves INT-1553

- Added the support for page custom event name
- generatePageCustomEventName utility -> Generates a custom event name for a page with given event handlebar
- Used the` handlebars.js` library for dynamic event template.
- useUserDefinedPageEventName - Toggle to decide whether to send a custom page event name
- userDefinedPageEventTemplate - Setting for event template

## Linear task (optional)

https://linear.app/rudderstack/issue/INT-1553/mixpanel-cloud-mode-make-page-and-screen-calls-naming-configurable

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

- **New Features**
	- Introduced the ability for users to define custom event names for page views in Mixpanel integration using templates.


ItsSudip (dismissed):

package.json
```diff
@@ -84,6 +84,7 @@
     "crypto-js": "4.2.0",
     "deep-object-diff": "1.1.9",
     "get-value": "3.0.1",
+    "handlebars": "^4.7.8",
```
aashishmalik, line 87: why we need a new package here? we can do with normal manipulation as I see size is increased with this import.

Gauravudia: Normal manipulation will not give the right output for template input. Down the line we have to update amplitude js sdk too to support this.

packages/analytics-js-integrations/__tests__/integrations/Mixpanel/util.test.js
```diff
@@ -474,12 +448,52 @@ describe('formatTraits', () => {
...
     });
     expect(result.setOnce).toEqual({
-      firstName: 'John'
+      firstName: 'John',
     });
   });
 });
 
+describe('generatePageCustomEventName', () => {
+  it('should generate a custom event name when userDefinedEventTemplate contains handlebars and message object is provided', () => {
```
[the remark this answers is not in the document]
Gauravudia: This testcase is covered here
`should return a default `Loaded a Page` event name when userDefinedEventTemplate contains invalid handlebars and message object is provided`

bardisg:
You should not need handlebars to crete event templates, nor any templating engine of that sorts.

Just write a method yourself

packages/analytics-js-integrations/__tests__/integrations/Mixpanel/util.test.js
```diff
@@ -474,12 +448,52 @@ describe('formatTraits', () => {
...
+      customField2: 'value2',
     });
     expect(result.setOnce).toEqual({
-      firstName: 'John'
+      firstName: 'John',
     });
   });
 });
 
+describe('generatePageCustomEventName', () => {
```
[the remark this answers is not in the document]
Gauravudia: added

packages/analytics-js-integrations/src/integrations/Mixpanel/browser.js
```diff
@@ -181,7 +184,24 @@ class Mixpanel {
...
+        logger.error(
+          'Event template is missing. Please provide a valid event template in `Page Event Name Template` setting',
+        );
+        return;
+      }
+      const eventName = generatePageCustomEventName(
+        rudderElement.message,
+        this.userDefinedPageEventTemplate,
+      );
+      window.mixpanel.track(eventName, properties);
```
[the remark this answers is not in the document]
Gauravudia: If name is not present in payload, we can just send event template without name placeholder {{ }},
Added a testcase for this in util.test.ts

packages/analytics-js-integrations/src/integrations/Mixpanel/util.js
```diff
@@ -241,21 +240,32 @@ const getConsolidatedPageCalls = config =>
...
-    const eventTemplate = Handlebars.compile(userDefinedEventTemplate);
-    return eventTemplate(message);
-  } catch (error) {
-    logger.error(`Error generating custom event name: ${error.message}`);
-    // Return a default or empty string as a fallback
-    return 'Loaded a Page';
+  let eventName = userDefinedEventTemplate
+    .replace('{{ category }}', message.properties?.category || '')
+    .trim();
+  eventName = eventName.replace('{{ name }}', message.name || '').trim();
```
saikumarrs, line 258:
```suggestion
  let eventName = userDefinedEventTemplate
    .replace('{{ category }}', message.properties?.category || '').replace('{{ name }}', message.name || '').trim();
```

packages/analytics-js-integrations/src/integrations/Mixpanel/util.js
```diff
@@ -241,21 +240,32 @@ const getConsolidatedPageCalls = config =>
...
+  // Remove any extra space between placeholders
+  eventName = eventName.replace(/\s{2,}/g, ' ');
```
saikumarrs, line 260: We must do this only if the placeholders are replaced with empty strings. Let's not mess up the user's event format.

packages/analytics-js-integrations/src/integrations/Mixpanel/util.js
```diff
@@ -241,21 +240,32 @@ const getConsolidatedPageCalls = config =>
...
+
+  // Check if any placeholders remain
+  if (eventName.includes('{{')) {
+    // Handle the case where either name or category is missing
+    eventName = eventName.replace(/{{\s*\w+\s*}}/g, '');
   }
```
saikumarrs, line 266: If we take responsibility for doing this, we should also handle other scenarios like malformed templates (orphan `{{`), etc., so let's skip it.

saikumarrs: The onus is completely on the user to configure the template correctly.

saikumarrs (changes requested):

packages/analytics-js-integrations/src/integrations/Mixpanel/util.js
Gauravudia: So, should we remove this step completely of removing extra space from template?

saikumarrs: We have to do it more intelligently. Otherwise, we could end up disrupting the template.

For a template like the one below (note extra spaces are already present),
`  Viewed {{ name }} {{ category }} page  some   keyword`


If `name` is unavailable, the template becomes,
`  Viewed  {{ category }} page  some   keyword`

the cleaned template should have looked like this,
`  Viewed {{ category }} page  some   keyword`


If `category` is unavailable, the template becomes,
`  Viewed  page  some   keyword`

the cleaned template should have looked like this,
`  Viewed page  some   keyword`


So, you can see that we merge the two whitespaces into 1 only if a single whitespace on either side surrounds the handlebars.

Gauravudia: Updated the regex.

saikumarrs: No, that change is not adequate. Please reach out to me so we can discuss this.

packages/analytics-js-integrations/src/integrations/Mixpanel/util.js
```diff
@@ -239,6 +239,38 @@ const getConsolidatedPageCalls = config =>
...
+ * const userDefinedEventTemplate = "Viewed {{ category }} {{ name }} Page";
+ * const message = {name: 'Home', properties: {category: 'Index'}};
+ * output: "Viewed Index Home Page"
+ */
+const generatePageCustomEventName = (message, userDefinedEventTemplate) => {
+  let eventName = userDefinedEventTemplate;
+
+  // Remove extra whitespace only if placeholders value are not present
+  if (!message.properties?.category) {
+    eventName = eventName.replace(/{{\s*category\s*}}\s+/g, '');
```
saikumarrs, line 259:
```suggestion
    eventName = eventName.replace(/\s{{\s*category\s*}}\s/g, ' ');
```

packages/analytics-js-integrations/src/integrations/Mixpanel/util.js
```diff
@@ -239,6 +239,39 @@ const getConsolidatedPageCalls = config =>
...
+
+  // find {{ name }} surrounded by whitespace characters and replace it with a single whitespace character
+  if (!isDefinedAndNotNull(message.name)) {
+    eventName = eventName.replace(/\s{{\s*name\s*}}\s/g, ' ');
+  }
+
+  // Replace placeholders with actual values
+  eventName = eventName
+    .replace('{{ category }}', message.properties?.category)
+    .replace('{{ name }}', message.name);
```
saikumarrs, line 270: Should also use the same regex format here?
`{{\s*category\s*}}` instead of `{{ category }}`

saikumarrs: Also, you can replace with actual values in the else blocks of the above if blocks.

if (property is defined) {
 replace the values
} else {
 check for surrounding whitespace and fuse them
}

saikumarrs (approved):

anantjain45823 (approved):

mihir-4116 (approved):

mihir-4116 (approved):

PR #1652, opened 2024-03-20
chore: remove duplicate config named eventWhiteList

ItsSudip:
## PR Description

Resolves INT-672
We are removing the duplicate config named eventWhiteList which is the same as the default SDK config named allowlist.

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

- **Refactor**
	- Removed event whitelisting logic from `TVSquared` integration, allowing all events to be processed without filtering.


packages/analytics-js-integrations/src/integrations/TVSquared/browser.js
```diff
@@ -21,7 +21,6 @@ class TVSquared {
     this.analytics = analytics;
     this.brandId = config.brandId;
     this.clientId = config.clientId;
-    this.eventWhiteList = config.eventWhiteList || [];
```
utsabc, line 24: Where are we storing allow list?

ItsSudip: The allow list is sdk level feature we are not storing it on destination specific browser.js files.

utsabc: ok

utsabc (approved):

krishna2020 (approved):

yashasvibajpai (approved):

MoumitaM (approved):

PR #1723, opened 2024-05-21
chore: fix github publish action

saikumarrs:
## PR Description

To avoid Nx considering all the projects affected for creating GitHub releases, as they usually involve changes to package.json and package-lock.json, we're explicitly ignoring them from Nx.

Also, all the project-specific dependencies have been moved to the corresponding package.json files. This was done to ensure any real dependency changes would result in releasing the package as it is a file within the project's scope.

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-1489/fix-the-publish-to-github-workflow-to-not-fail-when-a-package-does-not

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

- **Chores**
  - Updated GitHub workflow titles for better clarity in release pull requests.
  - Implemented a workaround to ignore `package-lock.json` and `package.json` files before publishing releases.
  - Restructured workflow steps in deployment to enhance deployment job referencing and parameter setting.
  - Updated URLs and project names for NPM packages in Slack notifications for better consistency and clarity.
  - Enhanced Slack message template for deployment notifications to include project name and JavaScript icon.
  - Added `REMOTE_MODULES_BASE_PATH` environment variable and modified project details in Slack notification for improved deployment information.

### DIFF
--- a/packages/analytics-v1.1/package.json
+++ b/packages/analytics-v1.1/package.json
@@ -76,12 +76,14 @@
     "url": "https://github.com/rudderlabs/rudder-sdk-js.git",
     "directory": "packages/analytics-v1.1"
   },
-  "dependencies": {
-    "component-emitter": "2.0.0",
-    "@segment/localstorage-retry": "1.3.0"
-  },
+  "dependencies": {},
   "devDependencies": {
-    "@rudderstack/analytics-js-common": "*"
+    "@rudderstack/analytics-js-common": "*",
+    "component-emitter": "2.0.0",
+    "@segment/localstorage-retry": "1.3.0",
+    "@lukeed/uuid": "2.0.1",
+    "get-value": "3.0.1",
+    "ramda": "0.30.0"
   },
   "overrides": {},
   "browserslist": {
--- a/.github/workflows/deploy-beta.yml
+++ b/.github/workflows/deploy-beta.yml
@@ -11,23 +11,17 @@ env:
   NODE_OPTIONS: "--no-warnings"
 
 jobs:
-  deploy:
-    name: Deploy BETA/BugBash Feature
+  get-deploy-inputs:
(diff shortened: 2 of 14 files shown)


packages/analytics-v1.1/package.json
```diff
@@ -76,12 +76,14 @@
...
-  },
+  "dependencies": {},
   "devDependencies": {
-    "@rudderstack/analytics-js-common": "*"
+    "@rudderstack/analytics-js-common": "*",
+    "component-emitter": "2.0.0",
+    "@segment/localstorage-retry": "1.3.0",
+    "@lukeed/uuid": "2.0.1",
+    "get-value": "3.0.1",
+    "ramda": "0.30.0"
```
MoumitaM, line 86: Haven't checked other packages but ramda, get-value, uuid, localstorage-retry these npm pkgs are dependencies, not devDependencies.

MoumitaM (changes requested):
You have removed all the dependencies from root package.json and added in devDependencies of respective packages. But this should be added in dependencies not in devDependencies.

packages/analytics-v1.1/package.json
saikumarrs: You are correct. I did that intentionally.

When we publish the v1.1 and v3 SDK NPM packages, the corresponding package.json will also be part of them as it is. If we define these dependencies as "prod", the users will be forced to install them. 
However, our final SDK is a rolled-up JavaScript bundle that does not need any external dependencies.
So, to prevent unnecessary dependency installations and carry over any potential vulnerabilities for our users, I marked them as "dev". Our development, bundling, and testing will have no effect because of this.

MoumitaM (dismissed):

PR #1802, opened 2024-07-23
feat(analytics-js-integrations): onboard ga4 v2 hybrid mode

utsabc:
## PR Description

Onboarding hybrid mode implementation for GA4 V2, this uses the existing GA4 implementation under the hood, with minor updates. The main change is that we are not going to execute the cookie migration logic if its GA4 V2

## Linear task (optional)

https://linear.app/rudderstack/issue/INT-2444/%5Bfeature%5D-onboard-ga4-v2-for-hybrid-mode

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


packages/analytics-js-integrations/src/integrations/GA4/browser.js
```diff
@@ -39,6 +39,7 @@ export default class GA4 {
     this.overrideClientAndSessionId = config.overrideClientAndSessionId || false;
     this.sdkBaseUrl =
       removeTrailingSlashes(config.sdkBaseUrl) || 'https://www.googletagmanager.com';
+    this.isExtendedGa4_V2 = config.isExtendedGa4_V2 || false;
```
saikumarrs, line 42: We're polluting GA4 implementation with the "v2" context. 
Ideally, the common code parts should be extracted into a separate module and version-specific logic should remain in individual integration code bases.

It does not need to be addressed now. So, please create a refactoring ticket and resolve it yourself.

saikumarrs (approved):
LGTM. Although, I have not reviewed any functional aspects of the new integration which I believe would be checked by other reviewers and already covered in E2E tests.

koladilip (approved):

packages/analytics-js-integrations/src/integrations/GA4/browser.js
krishna2020: Agree, need to bring loadScript outside for common code & have 2 different implementation for GA4 & GA4_v2

utsabc: The problem is the load script updates member variables of the parent class, just bringing load script wont work we need to figure out an alternative

krishna2020: @utsabc I see you tried to avoid code duplication but that should be fine instead of making it complex. `getClientDetails` method could also have been avoided in this class with little code duplication

utsabc: Yeah I will create backlog

PR #1856, opened 2024-09-18
fix: auto session tracking

MoumitaM:
## PR Description

This PR has changes to fix the issue, when session information exists in storage and autoTrack is set to false via loadOption, event payload had sessionId.

## Cross Browser Tests

- [x] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced session management logic to ensure correct behavior based on tracking options, preventing invalid session data usage.
	- Adjusted expected outcomes for session info under specific conditions.

- **Tests**
	- Improved test coverage for the `UserSessionManager` component, focusing on session behavior with `autoTrack` and `manualTrack` flags, including new scenarios for session data handling.

### DIFF
--- a/packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
+++ b/packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
@@ -203,6 +203,45 @@ describe('User session manager', () => {
       userSessionManager.syncStorageDataToState();
       expect(state.session.anonymousId.value).toBe('dummy-anonymousId');
     });
+    it('should not set sessionInfo if autoTrack is set to false in loadOption and sessionInfo exists in storage', () => {
+      const customData = {
+        rl_session: {
+          id: 1726655503445,
+          expiresAt: Date.now() + 60 * 1000,
+          timeout: 60000,
+          autoTrack: true,
+          sessionStart: false,
+        },
+      };
+      setDataInCookieStorage(customData);
+      state.loadOptions.value.sessions.autoTrack = false;
+      state.storage.entries.value = entriesWithOnlyCookieStorage;
+      userSessionManager.syncStorageDataToState();
+      expect(state.session.sessionInfo.value).toStrictEqual({});
+    });
+    it('should set sessionInfo if autoTrack is set to false in loadOption and sessionInfo exists in storage with manualTrack enabled', () => {
+      const customData = {
+        rl_session: {
+          id: 1726655503445,
+          expiresAt: Date.now() + 60 * 1000,
+          timeout: 60000,
+          manualTrack: true,
+          sessionStart: false,
+        },
+      };
+      setDataInCookieStorage(customData);
+      state.loadOptions.value.sessions.autoTrack = false;
+      state.storage.entries.value = entriesWithOnlyCookieStorage;
+      userSessionManager.syncStorageDataToState();
+      expect(state.session.sessionInfo.value).toStrictEqual({
+        id: 1726655503445,
+        expiresAt: expect.any(Number),
+        timeout: 60000,
+        manualTrack: true,
+        autoTrack: false,
+        sessionStart: false,
+      });
+    });
   });
 
   describe('init', () => {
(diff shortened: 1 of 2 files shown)


packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
```diff
@@ -203,6 +203,66 @@ describe('User session manager', () => {
       userSessionManager.syncStorageDataToState();
       expect(state.session.anonymousId.value).toBe('dummy-anonymousId');
     });
+    it('should not set sessionInfo if autoTrack is set to false in loadOption and sessionInfo exists in storage', () => {
```
saikumarrs, line 206: If we were to run this test case without the code changes in this PR, would it fail?

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
```diff
@@ -142,9 +142,11 @@ class UserSessionManager implements IUserSessionManager {
       };
     }
 
-    state.session.sessionInfo.value = this.isPersistenceEnabledForStorageEntry('sessionInfo')
-      ? (sessionInfo as SessionInfo)
-      : DEFAULT_USER_SESSION_VALUES.sessionInfo;
+    state.session.sessionInfo.value =
```
saikumarrs, line 145: At this point, the `sessionInfo` variable has already taken into account the configured load options in Ln 139 and 140. 
The load option values take precedence.

Ln 140 says that auto-tracking should be enabled only if the load option flag is set and in the storage manual tracking is not configured.

Are we trying to solve a different problem where auto tracking is configured to be off so we do not persist anything in the storage (by using the default value `{}`)?

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
```diff
@@ -142,9 +142,11 @@ class UserSessionManager implements IUserSessionManager {
       };
     }
```
saikumarrs, line 143: To simplify the logic further, we should add an else case here and assign `sessionInfo` to the default value.

In the next step, simply assign `sessionInfo` to the state.

saikumarrs: It'll avoid calling `isPersistenceEnabledForStorage` twice.

saikumarrs (changes requested):
It is not clear the bug we're trying to solve in this PR. Let's discuss this offline.

packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
MoumitaM: Yes

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
MoumitaM: The primary problem is with existing session info in the storage, when autoTrack is set to false via loadOptions, we are still seeing sessionId in the event payload. The reason behind this is we are not clearing the sessionInfo.
So, by setting it to default value we are resolving both the problems.

saikumarrs: At what point in the code that bug is introduced? 
The logic seems clear as I explained above.

MoumitaM: Not sure

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
MoumitaM: Refactored

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
saikumarrs: What do you mean by not sure? Have you not debugged the issue line by line using breakpoints?

saikumarrs (approved):

PR #1876, opened 2024-10-08
feat: track time spent on a page

MoumitaM:
## PR Description

New feature:
- Track page lifecycle events(Page Loaded, Page Unloaded) with configurable load options
- Calculate the time spent on a page.

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [x] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Release Notes

- **New Features**
  - Introduced page lifecycle tracking with new options for tracking page load and unload events.
  - Enhanced the analytics interface to include unique visit identifiers and page load timestamps.
  - New state management for page lifecycle events to improve tracking capabilities.
  - Expanded logging capabilities with a new log level for post-load events.
  - Introduced a new commented-out property for future tracking options in the configuration.
  - Updated documentation to clarify usage of the RudderStack JavaScript SDK and variable naming conventions.

- **Bug Fixes**
  - Enhanced validation for tracking options to ensure proper configuration.

- **Chores**
  - Updated size limits for various analytics components to accommodate recent changes.
  - Added new types and enums to improve analytics functionality and state management.
  - Enhanced tests for the `onPageLeave` function to ensure correct behavior under various conditions.


packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -133,6 +141,74 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
...
+   * @param loadOptions
+   * @returns
+   */
+  // eslint-disable-next-line class-methods-use-this
+  trackPageLifecycleEvents(
+    preloadedEventsArray: PreloadedEventCall[],
+    loadOptions?: Partial<LoadOptions>,
+  ) {
+    const { trackPageLifecycle, useBeacon } = loadOptions ?? {};
+    const { events = [], enabled = false, options = {} } = trackPageLifecycle ?? {};
```
saikumarrs, line 156: A better option would be to initialize `events` with the default value of both the events here.

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -133,6 +141,74 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
...
+      const visitId = generateUUID();
+      const pageLoadedTimestamp = Date.now();
+      if (events.length === 0 || events.includes(PageLifecycleEvents.PAGELOADED)) {
+        preloadedEventsArray.unshift([
+          'track',
+          PageLifecycleEvents.PAGELOADED,
+          { visitId },
+          {
+            originalTimestamp: new Date(pageLoadedTimestamp).toISOString(),
+            ...options,
```
saikumarrs, line 167: We should not let the user override the default `originalTimestamp` parameter. It'll mess up other things for us.
So, let's promote `...options` ahead of `originalTimestamp`.

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -133,6 +141,74 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
...
+          },
+        ]);
+      }
+      if (events.length === 0 || events.includes(PageLifecycleEvents.PAGEUNLOADED)) {
+        if (useBeacon === true) {
+          // Register the page unloaded lifecycle event listeners
+          onPageLeave((isAccessible: boolean) => {
+            if (isAccessible === false) {
+              const visitDuration = Date.now() - pageLoadedTimestamp;
```
saikumarrs, line 176: We should store the current timestamp here again and override it as the original timestamp through the options as before.
Even if the user calculates the duration again from the timestamps in the events, it should perfectly match the `visitDuration` field.

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -133,6 +141,74 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
...
+              if (!state.lifecycle.loaded.value) {
```
saikumarrs, line 177: It is good that you've thought about this scenario. If the SDK is not in the loaded state, there is no point in queuing the event in the preloaded buffer. The SDK will not have a chance to even put it into local storage, let alone send it.

However, similar to this approach, when SDK is loaded, we need to promote this event ahead of all the stored events. Invoking the `track` API will by default add this event to the end of the queued events.
The last batch flushed may not include this event. We need to maximize the chance.

packages/analytics-js/src/constants/logMessages.ts
```diff
@@ -257,6 +257,9 @@ const INVALID_POLYFILL_URL_WARNING = (
 const BAD_COOKIES_WARNING = (key: string) =>
   `The cookie data for ${key} seems to be encrypted using SDK versions < v3. The data is dropped. This can potentially stem from using SDK versions < v3 on other sites or web pages that can share cookies with this webpage. We recommend using the same SDK (v3) version everywhere or avoid disabling the storage data migration.`;
 
+const PAGE_UNLOAD_ON_BEACON_DISABLED_WARNING = (): string =>
+  `Page Unloaded event is only supported for Beacon transport mechanism. Please set useBeacon to true to enable this feature.`;
```
saikumarrs, line 261: We should include context similar to other messages.

packages/analytics-js/src/constants/logMessages.ts
```diff
@@ -257,6 +257,9 @@ const INVALID_POLYFILL_URL_WARNING = (
 const BAD_COOKIES_WARNING = (key: string) =>
   `The cookie data for ${key} seems to be encrypted using SDK versions < v3. The data is dropped. This can potentially stem from using SDK versions < v3 on other sites or web pages that can share cookies with this webpage. We recommend using the same SDK (v3) version everywhere or avoid disabling the storage data migration.`;
 
+const PAGE_UNLOAD_ON_BEACON_DISABLED_WARNING = (): string =>
```
saikumarrs, line 260: If it is a static string value, it should not be a function. Let's define it as a string constant.

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -155,6 +231,8 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
 
     // Process load method if present in the buffered requests
     if (loadEvent.length > 0) {
+      // Track page loaded lifecycle event if enabled
+      this.trackPageLifecycleEvents(preloadedEventsArray, loadEvent[3]);
```
saikumarrs, line 235: The load options normalization, etc., will happen in the load API of the Analytics class, whereas we're relying on that data (in the track life cycle function) before the load is even initiated.
Please double-check it.

saikumarrs (changes requested):

packages/analytics-js/src/app/RudderAnalytics.ts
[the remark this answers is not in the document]
MoumitaM: But we need load options to understand whether the feature is enabled. Without knowing that should we create these events?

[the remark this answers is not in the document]
saikumarrs: It looks like we need to generate a visitor ID and timestamp irrespective of this feature's status. So, we should generate these values as early as possible to track the page's duration effectively.

However, the events should be generated only after the load API is invoked.

packages/analytics-js-common/src/types/LoadOptions.ts
```diff
@@ -108,6 +108,17 @@ export type PreConsentOptions = {
   events?: PreConsentEventsOptions;
 };
 
+export enum PageLifecycleEvents {
+  PAGELOADED = 'Page Loaded',
+  PAGEUNLOADED = 'Page Unloaded',
```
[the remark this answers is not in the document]
MoumitaM: Addressed

packages/analytics-js/src/constants/logMessages.ts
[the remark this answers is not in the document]
MoumitaM: Addressed

packages/analytics-js/src/app/RudderAnalytics.ts
MoumitaM: Addressed

packages/analytics-js/src/app/RudderAnalytics.ts
MoumitaM: Addressed

packages/analytics-js/src/app/RudderAnalytics.ts
MoumitaM: Addressed

packages/analytics-js/src/constants/logMessages.ts
MoumitaM: Converted to a static string

packages/analytics-js/src/constants/logMessages.ts
MoumitaM: Addressed

packages/analytics-js/src/constants/logMessages.ts
saikumarrs: It is not enough. The context is needed.
Something similar to this,

```
${context}${LOG_CONTEXT_SEPARATOR}Page Unloaded event can only be tracked when the Beacon transport is active. Please enable "useBeacon" load API option.
```

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -80,7 +88,7 @@
     this.consent = this.consent.bind(this);
 
     RudderAnalytics.globalSingleton = this;
-
+    this.logger?.setMinLogLevel('WARN');
```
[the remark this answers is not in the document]
MoumitaM: loadOption is determined at a later stage, unfortunately, not applicable here.

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -155,6 +230,8 @@
 
     // Process load method if present in the buffered requests
     if (loadEvent.length > 0) {
+      // Track page loaded lifecycle event if enabled
+      this.trackPageLifecycleEvents(preloadedEventsArray, loadEvent[3]);
```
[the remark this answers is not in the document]
MoumitaM: Not required

packages/analytics-js/src/app/RudderAnalytics.ts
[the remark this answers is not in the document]
MoumitaM: Addressed

packages/analytics-js/src/constants/logMessages.ts
MoumitaM: Addressed

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -133,6 +141,73 @@
...
+            );
+          }
+        });
+      } else {
+        // throw warning if beacon is disabled
+        this.logger.warn(PAGE_UNLOAD_ON_BEACON_DISABLED_WARNING);
+      }
+    }
+    setExposedGlobal(GLOBAL_PRELOAD_BUFFER, clone(preloadedEventsArray));
+  }
```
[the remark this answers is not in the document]
MoumitaM: This implementation suggestion has multiple flaws and doesn't benefit much.

packages/analytics-js-common/src/types/IRudderAnalytics.ts
```diff
@@ -76,6 +76,8 @@ export interface IRudderAnalytics<T = any> {
   analyticsInstances: Record<string, T>;
   defaultAnalyticsKey: string;
   logger: ILogger;
+  visitId: string;
+  pageLoadedTimestamp: number;
```
saikumarrs, line 80: We can store it in the state, right?

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -133,6 +141,74 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
...
+            }
+          });
+        } else {
+          // throw warning if beacon is disabled
+          this.logger.warn(PAGE_UNLOAD_ON_BEACON_DISABLED_WARNING());
+        }
+      }
+      setExposedGlobal(GLOBAL_PRELOAD_BUFFER, clone(preloadedEventsArray));
+    }
+  }
```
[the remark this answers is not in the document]
MoumitaM: Outdated

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -80,7 +90,7 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
     this.consent = this.consent.bind(this);
 
     RudderAnalytics.globalSingleton = this;
-
+    this.logger?.setMinLogLevel('WARN');
```
saikumarrs, line 93: We might be better off changing the default log level to the lowest level instead of setting it here. So, we can capture all kinds of messages.

Again, after the SDK loads, we should set the log level to 'ERROR' (default value) if the user doesn't configure it.

Essentially, it's like having two default values before and after SDK loads.

packages/analytics-js/src/app/RudderAnalytics.ts
MoumitaM: Addressed

packages/analytics-js-common/src/types/IRudderAnalytics.ts
MoumitaM: Yes we can, but right now it is not needed. If requires I will add it.

saikumarrs: If we add it to the state, it'll help us with debugging as well. Please think about it.

MoumitaM: Addressed

packages/analytics-js/src/app/RudderAnalytics.ts
MoumitaM: Addressed

packages/analytics-js/src/app/RudderAnalytics.ts
MoumitaM: > It is good that you've thought about this scenario. If the SDK is not in the loaded state, there is no point in queuing the event in the preloaded buffer. The SDK will not have a chance to even put it into local storage, let alone send it.

Addressed

> However, similar to this approach, when SDK is loaded, we need to promote this event ahead of all the stored events. Invoking the track API will by default add this event to the end of the queued events.
The last batch flushed may not include this event. We need to maximize the chance.

This is not needed, I have tested this scenario by reducing the size limit.

packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -127,9 +127,7 @@ class Analytics implements IAnalytics {
     });
 
     // set log level as early as possible
-    if (state.loadOptions.value.logLevel) {
-      this.logger?.setMinLogLevel(state.loadOptions.value.logLevel);
-    }
+    this.logger?.setMinLogLevel(state.loadOptions.value.logLevel ?? 'ERROR');
```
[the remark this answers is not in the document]
MoumitaM: Not required

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -149,31 +153,34 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
...
+    const pageLoadedTimestamp = state.pageLifecycle.pageLoadedTimestamp.value as number;
+
     if (!enabled) {
       return;
     }
 
+    const preloadedEventsArray = Array.isArray((globalThis as typeof window).rudderanalytics)
+      ? ((globalThis as typeof window).rudderanalytics as unknown as PreloadedEventCall[])
+      : ([] as PreloadedEventCall[]);
+
```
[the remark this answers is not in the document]
MoumitaM: Addressed

packages/analytics-js/src/components/core/Analytics.ts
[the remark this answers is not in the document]
MoumitaM: Addressed

packages/analytics-js/src/app/RudderAnalytics.ts
MoumitaM: Please resolve this

saikumarrs: In the above scenario, after testing, it was realized that the events left in the buffer were automatically cleared by the flush handler registered on page leave.
So, effectively all the events were flushed. There is no need to jumble the order of the events.

saikumarrs (commented):
LGTM

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -129,18 +140,95 @@
...
+    } = trackPageLifecycle ?? {};
+
+    const visitId = state.pageLifecycle.visitId.value;
+    const pageLoadedTimestamp = state.pageLifecycle.pageLoadedTimestamp.value as number;
+
+    if (!enabled) {
+      return;
+    }
+
+    const preloadedEventsArray = this.getPreloadedEvents();
```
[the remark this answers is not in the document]
MoumitaM: Added unit tests.

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -129,18 +140,95 @@
...
+            );
+          }
+        });
+      } else {
+        // throw warning if beacon is disabled
+        this.logger.warn(PAGE_UNLOAD_ON_BEACON_DISABLED_WARNING(RS_APP));
+      }
+    }
+    setExposedGlobal(GLOBAL_PRELOAD_BUFFER, clone(preloadedEventsArray));
+  }
```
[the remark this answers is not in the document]
MoumitaM: Addressed

packages/analytics-js/src/services/Logger/Logger.ts
```diff
@@ -11,7 +11,8 @@ const LOG_LEVEL_MAP: Record<LogLevel, number> = {
   NONE: 5,
 };
 
-const DEFAULT_LOG_LEVEL = 'ERROR';
+const DEFAULT_LOG_LEVEL = 'LOG';
+const POST_LOAD_LOG_LEVEL = 'ERROR';
```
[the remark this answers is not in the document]
MoumitaM: Not required

packages/analytics-js/src/services/Logger/Logger.ts
```diff
@@ -11,7 +11,7 @@ const LOG_LEVEL_MAP: Record<LogLevel, number> = {
   NONE: 5,
 };
 
-const DEFAULT_LOG_LEVEL = 'ERROR';
+const DEFAULT_LOG_LEVEL = 'LOG';
```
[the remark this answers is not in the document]
MoumitaM: Yes the change is intentional

packages/analytics-js-common/src/types/LoadOptions.ts
```diff
@@ -108,6 +108,17 @@
...
+}
```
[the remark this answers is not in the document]
MoumitaM: Added unit tests

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -133,6 +141,74 @@
...
+            }
+          });
+        } else {
+          this.logger.warn(PAGE_UNLOAD_ON_BEACON_DISABLED_WARNING());
+        }
+      }
+      setExposedGlobal(GLOBAL_PRELOAD_BUFFER, clone(preloadedEventsArray));
+    }
+  }
+
```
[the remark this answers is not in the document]
MoumitaM: Added unit tests

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -129,18 +140,100 @@
...
+            );
+          }
+        });
+      } else {
+        // throw warning if beacon is disabled
+        this.logger.warn(PAGE_UNLOAD_ON_BEACON_DISABLED_WARNING(RS_APP));
+      }
+    }
+    setExposedGlobal(GLOBAL_PRELOAD_BUFFER, clone(preloadedEventsArray));
+  }
```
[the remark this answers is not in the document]
MoumitaM: Unit tests added

packages/analytics-js/__tests__/app/RudderAnalytics.test.ts
```diff
@@ -19,7 +19,7 @@ describe('Core - Rudder Analytics Facade', () => {
     (window as any).rudderanalytics = [
       ['track'],
       ['consent', { sendPageEvent: true }],
-      ['load', { option1: true }],
+      ['load', 'dummyWriteKey', 'dummyDataPLaneUrl', { option1: true }],
```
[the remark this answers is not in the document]
MoumitaM: Addressed

packages/analytics-js/__tests__/app/RudderAnalytics.test.ts
```diff
@@ -283,4 +283,138 @@
...
+    (globalThis as typeof window).rudderanalytics = undefined;
+    const result = rudderAnalyticsInstance.getPreloadedEvents();
+    expect(result).toEqual([]);
+  });
+
+  it('should return buffered events array when globalThis.rudderanalytics is an array', () => {
+    const bufferedEvents = [
+      ['track'],
+      ['consent', { sendPageEvent: true }],
+      ['load', 'dummyWriteKey', 'dummyDataPLaneUrl', { option1: true }],
```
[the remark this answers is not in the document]
MoumitaM: Addressed

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -165,62 +165,102 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
...
+      this.trackPageLoadedEvent(events, options, preloadedEventsArray);
+      this.setupPageUnloadTracking(events, useBeacon, options);
     }
 
-    const preloadedEventsArray = this.getPreloadedEvents();
+    // The array will be mutated in the below method
+    promotePreloadedConsentEventsToTop(preloadedEventsArray);
 
-    // track page loaded event
+    setExposedGlobal(GLOBAL_PRELOAD_BUFFER, clone(preloadedEventsArray));
```
saikumarrs, line 190: Let's see if we could move them out of this function. They seem to be misfit here.

packages/analytics-js-common/src/types/ApplicationState.ts
```diff
@@ -83,6 +83,10 @@ export type LifecycleState = {
   dataPlaneUrl: Signal<string | undefined>;
 };
 
+export type AutoTrackState = {
```
saikumarrs, line 86: Let's add the `enabled` status of auto tracking and page life cycle to the state.

packages/analytics-js/src/state/slices/loadOptions.ts
```diff
@@ -44,9 +44,12 @@ const defaultLoadOptions: LoadOptions = {
   },
   sendAdblockPageOptions: {},
   useServerSideCookies: false,
-  trackPageLifecycle: {
+  autoTrack: {
```
saikumarrs, line 47: For auto-tracking options, we can omit the default values as they don't go through the normalization.

saikumarrs (changes requested):

packages/analytics-js/src/app/RudderAnalytics.ts
MoumitaM: Addressed

packages/analytics-js-common/src/types/ApplicationState.ts
MoumitaM: Addressed

packages/analytics-js/src/state/slices/loadOptions.ts
MoumitaM: Addressed

saikumarrs (dismissed):

packages/analytics-js-common/src/utilities/page.ts
```diff
@@ -12,6 +12,12 @@
     pageLeft = true;
 
     callback(isAccessible);
+
+    // Reset pageLeft on the next tick
+    // to ensure callback executes for other listeners
+    setTimeout(() => {
+      pageLeft = false;
+    }, 0);
```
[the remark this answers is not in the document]
MoumitaM: Addressed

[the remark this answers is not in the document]
MoumitaM: Addressed

packages/analytics-js-common/src/utilities/page.ts
```diff
@@ -12,6 +12,12 @@ const onPageLeave = (callback: (isAccessible: boolean) => void) => {
     pageLeft = true;
 
     callback(isAccessible);
+
+    // Reset pageLeft on the next tick
+    // to ensure callback executes for other listeners
```
saikumarrs, line 17:
```suggestion
    // to ensure callback executes for other listeners
    //when closing an inactive browser tab.
```

packages/analytics-js-common/src/utilities/page.ts
saikumarrs, line 1: Please make sure the new code is covered a 100%. It is a critical piece.

packages/analytics-js-common/__tests__/utilities/page.test.ts
```diff
@@ -182,4 +182,21 @@ describe('onPageLeave', () => {
...
+    dispatchDocumentEvent('visibilitychange');
+
+    dispatchWindowEvent('beforeunload');
+
+    expect(evCallback).toHaveBeenCalledTimes(1);
+    expect(evCallback).toHaveBeenNthCalledWith(1, true);
+    setTimeout(() => {
+      expect(evCallback).toHaveBeenCalledTimes(1);
+      expect(evCallback).toHaveBeenNthCalledWith(1, false);
+    }, 0);
```
saikumarrs, line 200: These assertions will never be executed by Jest as they're are part of the asynchronous code block. You should use the `done` function in this unit test function.

packages/analytics-js-common/__tests__/utilities/page.test.ts
```diff
@@ -182,4 +182,21 @@ describe('onPageLeave', () => {
...
+    setVisibilityState('hidden');
+    dispatchDocumentEvent('visibilitychange');
+
+    dispatchWindowEvent('beforeunload');
+
+    expect(evCallback).toHaveBeenCalledTimes(1);
+    expect(evCallback).toHaveBeenNthCalledWith(1, true);
+    setTimeout(() => {
+      expect(evCallback).toHaveBeenCalledTimes(1);
+      expect(evCallback).toHaveBeenNthCalledWith(1, false);
```
saikumarrs, line 199: I think you meant to check if the callback was called again. But it doesn't work like that.
The mock accumulates the number of times it was called until it is reset.

You have two options:
- Reset the mock so that the invocation counters are cleared
- Check that it is called 2 times and verify the values for second invocation.

saikumarrs (changes requested):

packages/analytics-js-common/src/utilities/page.ts
MoumitaM: Addressed

packages/analytics-js-common/__tests__/utilities/page.test.ts
MoumitaM: Addressed

packages/analytics-js-common/__tests__/utilities/page.test.ts
MoumitaM: Addressed

packages/analytics-js-common/src/utilities/page.ts
MoumitaM: The calculation is not entirely correct please check the details. It Should be around 100%.

saikumarrs: The changes you made in this file are critical . So, we should aim for 100% code coverage.
SonarCloud quality gate is a general limit. It shouldn't stop us from covering more code with tests wherever possible.

Anyway, with the recent changes, this is addressed automatically.

MoumitaM: It is 100%

saikumarrs (approved):

PR #1904, opened 2024-10-28
fix: move page visit id to context

saikumarrs:
## PR Description

Page visit ID has been moved to context. It is attached to all the events.
Also, the global auto-tracking enabled status is set if at least one of the components are enabled.

Additionally, cleaned up some code in the src and test suites.

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

### DIFF
--- a/packages/analytics-js/src/app/RudderAnalytics.ts
+++ b/packages/analytics-js/src/app/RudderAnalytics.ts
@@ -22,6 +22,7 @@ import { isString } from '@rudderstack/analytics-js-common/utilities/checks';
 import type { IdentifyTraits } from '@rudderstack/analytics-js-common/types/traits';
 import { generateUUID } from '@rudderstack/analytics-js-common/utilities/uuId';
 import { onPageLeave } from '@rudderstack/analytics-js-common/utilities/page';
+import { getFormattedTimestamp } from '@rudderstack/analytics-js-common/utilities/timestamp';
 import { GLOBAL_PRELOAD_BUFFER } from '../constants/app';
 import {
   getPreloadedLoadEvent,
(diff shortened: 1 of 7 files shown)


packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -181,16 +182,20 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
...
-      enabled = autoTrackEnabled,
+      enabled: pageLifecycleEnabled = autoTrackEnabled,
       options = autoTrackOptions,
     } = pageLifecycle ?? {};
 
-    state.autoTrack.enabled.value = autoTrackEnabled;
-    state.autoTrack.pageLifecycle.enabled.value = enabled;
+    // Set the autoTrack enabled state
+    // if at least one of the autoTrack options is enabled
+    state.autoTrack.enabled.value = autoTrackEnabled || pageLifecycleEnabled;
```
MoumitaM, line 194: We shouldn't do this. In the future when we will have more tracking options here, those will be enabled unknowingly.

MoumitaM (changes requested):

packages/analytics-js/src/app/RudderAnalytics.ts
saikumarrs: If the page life cycle is automatically tracking, it means auto tracking is enabled in the SDK, in some or the other form.

In the future, when more components are auto-tracked, they all have to be processed before setting this flag.
Anyway, in that scenario, we have to handle the whole functionality more elegantly.

Should I add a note in the code to avoid any risk?

MoumitaM: Yes, let's add a note

MoumitaM (approved):

PR #1956, opened 2024-12-03
feat: lock plugins and integrations version by default

saikumarrs:
## PR Description

I've deprecated the `lockIntegrationsVersion` and `lockPluginsVersion` load API options as the SDK will load the integrations and plugins from the versioned directories by default now.

Even in the case of when user provides the paths for these components, the major version (`/v3`) will be replaced with the exact version of the core SDK.

Additionally, the core SDK script tag will be identified using the dataset attribute first before using the fallback option.

The deployment workflow has been updated with the following changes:
- No longer publish integrations and plugin artefacts to the default production path (`/v3`). They are published only to the versioned directories.
- Copy SDK dependencies ahead of the core artefacts.
- The HTML files to list the integrations and plugins are regenerated for the versioned directories.

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-2708/namespace-the-dynamically-loaded-sdk-artifacts-with-a-release-tag

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


packages/analytics-js/src/components/configManager/util/cdnPaths.ts
```diff
@@ -18,27 +19,39 @@ const getSDKComponentBaseURL = (
   lockVersion: boolean,
   customURL?: string,
 ) => {
-  let sdkComponentURL = '';
-
+  let sdkComponentBaseURL;
```
arpl, line 22: This can be simplified. For more than 3 arguments it's a good practice to use object so the call site is clear.

Having said that `baseURL` can be renamed to `fallBack` or `defaultURL`

And let's rename `customURL` to `loadOptionsURL`.

arpl (approved):

.github/workflows/deploy.yml
```diff
@@ -135,106 +160,82 @@ jobs:
...
+          aws s3 cp ${{ env.CORE_ARTIFACTS_BASE_PATH }}/legacy/iife/ $s3_path_prefix/legacy/ $copy_args
+          aws s3 cp ${{ env.CORE_ARTIFACTS_BASE_PATH }}/modern/iife/ $s3_path_prefix/modern/ $copy_args
 
-          aws s3 cp $integration_sdks_path_prefix/legacy/js-integrations/ $s3_path_prefix/legacy/js-integrations/ $copy_recursive_args
-          aws s3 cp $plugins_path_prefix/modern/plugins/ $s3_path_prefix/modern/plugins/ $copy_recursive_args
-          aws s3 cp $integration_sdks_path_prefix/modern/js-integrations/ $s3_path_prefix/modern/js-integrations/ $copy_recursive_args
+      - name: Invalidate CloudFront cache for all the SDK artifacts (versioned directory)
+        if: ${{ inputs.environment == 'production' }}
+        run: |
+          invalidation_id=$(AWS_MAX_ATTEMPTS=10 aws cloudfront create-invalidation --distribution-id ${{ secrets.AWS_CF_DISTRIBUTION_ID }} --paths "/${{ env.CURRENT_VERSION_VALUE }}/*" --query "Invalidation.Id" --output text)
```
arpl, line 199: We don't need to invalidate these since no-one knows about them yet.

Even if we discover a defect and we should release the fix under a new version.

In a rollback scenario we will need to invalidate the rolled back version paths.

ChryssaAliferi (approved):

.github/workflows/deploy.yml
saikumarrs: Done.

packages/analytics-js/src/components/configManager/util/cdnPaths.ts
saikumarrs: Done.

PR #1994, opened 2025-01-07
fix: braze anonymousId tracking with alias details

utsabc:
### Anonymous User Identification

- Added automatic alias creation by setting the RudderStack anonymous_id as an alias_id in Braze when the SDK initializes
- Aligned device mode behavior with cloud mode implementation for consistent user tracking
- Fixed incorrect usage of Braze's change_user API for anonymous user tracking

### Profile Management

- Implemented proper linking between device_id and anonymous_id to maintain a single user profile
- Removed redundant user_id assignments that were causing profile duplication
- Streamlined page tracking implementation to maintain consistent user identification

We achieve the above by making the following Key Changes
- We are setting the anonymous_id as braze alias label in the isReady call back
- We are calling changeUser only in case of identify [change user api doc](https://js.appboycdn.com/web-sdk/latest/doc/modules/braze.html#changeuser) 

[Linear Ticket](https://linear.app/rudderstack/issue/INT-2903/braze-aman-are-seeing-anonymous-events-creating-identified-users-in)

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

- **New Features**
  - Enhanced Braze integration with improved event tracking mechanisms.

- **Bug Fixes**
  - Streamlined event logging process for more reliable tracking.

- **Refactor**
  - Simplified tracking and page view methods to improve code clarity and performance.
  - Updated handlePurchase function to operate without user ID, focusing on logging purchases.
  - Introduced a mock for the `addAlias` method in the test suite to enhance testing coverage.


packages/analytics-js-integrations/__tests__/integrations/Braze/browser.test.js
```diff
@@ -151,6 +152,30 @@ describe('isLoaded', () => {
...
+
+    const braze = new Braze(config, analytics, destinationInfo);
+    braze.init();
+    // mock the window.braze
+    mockBrazeSDK();
+    
+    // mock true for isLoaded
+    jest.spyOn(braze, 'isLoaded').mockReturnValue(true);
+
+    jest.spyOn(window.braze, 'addAlias');
```
saikumarrs, line 173: It's already a mock function so there is no need to spy on it again.

packages/analytics-js-integrations/__tests__/integrations/Braze/browser.test.js
```diff
@@ -734,11 +753,14 @@ describe('track', () => {
       },
     };
 
-    jest.spyOn(window.braze, 'changeUser');
     braze.track(rudderElement);
 
     // Expect the necessary Braze methods to be called with the correct values
-    expect(window.braze.changeUser).toHaveBeenCalledWith('anon123');
+    expect(window.braze.logCustomEvent).toHaveBeenCalledWith('Product Reviewed', {
```
saikumarrs, line 759: Nitpick: You might also want to ensure that it is called exactly once.

packages/analytics-js-integrations/src/integrations/Braze/browser.js
```diff
@@ -75,7 +75,17 @@ class Braze {
...
   isReady() {
-    return this.isLoaded();
+    if (this.isLoaded()) {
+      try {
+        const anonymousId = this.analytics.getAnonymousId();
+        window.braze.getUser().addAlias(anonymousId, 'rudder_id');
+        return true;
+      } catch (error) {
+        logger.error(`Error in isReady - ${stringifyWithoutCircularV1(error, true)}`);
+        return false;
```
saikumarrs, line 85: If `false` is returned, the core SDK will repeatedly attempt the `isReady` check, leading to multiple error statements on the console.

Do we want to handle this better?

packages/analytics-js-integrations/src/integrations/Braze/browser.js
```diff
@@ -75,7 +75,17 @@ class Braze {
   }
 
   isReady() {
-    return this.isLoaded();
+    if (this.isLoaded()) {
+      try {
+        const anonymousId = this.analytics.getAnonymousId();
+        window.braze.getUser().addAlias(anonymousId, 'rudder_id');
```
saikumarrs, line 81: Can we use a more appropriate property name here? Like `rudderstack_id`

utsabc: We need to maintain it keep consistency for cloud mode and hybrid mode
ref: https://github.com/rudderlabs/rudder-transformer/blob/3c6e0476e7537574ca150e494096c2ecfde93008/src/v0/destinations/braze/util.js#L609

packages/analytics-js-integrations/src/integrations/Braze/browser.js
utsabc: Any suggestion?

saikumarrs: You can maintain the state of this operation in the class to log just once and then log again if it gets successful.

Moreover, the API returns a [boolean](https://js.appboycdn.com/web-sdk/latest/doc/classes/braze.user.html#addalias) which we're ignoring. Apart from the error handling, we should also use the return value to maintain the state.

Also, do we want to consider the destination as not ready if alias couldn't be set?

packages/analytics-js-integrations/src/integrations/Braze/browser.js
krishna2020, line 1: should the switch to userId as part of identify be limited only to hybrid mode ? https://github.com/rudderlabs/rudder-sdk-js/pull/1994/files#diff-553db15b80ab82e1e3d027f9108f29df290199053909e8b9295de14fec5d3366L118

utsabc: This was existing logic itself

packages/analytics-js-integrations/src/integrations/Braze/browser.js
utsabc: Without setting the `alias` we cannot set the `anonymousId` to the profile which is required for
- Keeping these profiles consistent like cloud mode
- Stitch profiles for hybrid mode

utsabc: I didn't get state part on every page refresh this entire class script will get executed how will be the state maintained in that case? Or are talking of using some session storage kind of state?

utsabc: I have updated the code can you check once

packages/analytics-js-integrations/src/integrations/Braze/browser.js
```diff
@@ -74,18 +85,38 @@ class Braze {
...
-      try {
-        const anonymousId = this.analytics.getAnonymousId();
-        window.braze.getUser().addAlias(anonymousId, 'rudder_id');
-        return true;
-      } catch (error) {
-        logger.error(`Error in isReady - ${stringifyWithoutCircularV1(error, true)}`);
+  setUserAlias() {
+    try {
+      const anonymousId = this.analytics.getAnonymousId();
+      if (!anonymousId) {
```
saikumarrs, line 91: You can skip this case as anon ID will always be available.

saikumarrs (approved):

packages/analytics-js-integrations/src/integrations/Braze/browser.js
utsabc: Keeping it for consistency :)

packages/analytics-js-integrations/src/integrations/Braze/browser.js
```diff
@@ -74,8 +85,38 @@ class Braze {
     return window.brazeQueue === null;
   }
 
+  setUserAlias() {
```
krishna2020, line 88: there are 3 log statements here but only 1 of them works as first invocation of log message will set hasLoggedErrorForAlias to true

utsabc: Yeah at a time one these three scenarios can happen - so will log for that specific scenario

krishna2020 (approved):

PR #2007, opened 2025-01-20
feat: make error handler and logger mandatory in all components

saikumarrs:
## PR Description

Up until now error handler and logger instances are optional properties. However, in the Analytics instance they are defined by default and passed along. 
So, to simply the code, now we've updated all the components to reflect the status.

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-2826/move-error-reporting-functionality-to-the-core-sdk

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


ChryssaAliferi (approved):
🙌

arpl (approved):
looks find although the PR is sprinkled with changes other than removing optionality from logger and errorHandler.

PR #2056, opened 2025-02-21
fix: avoid premature execution before sdk initialization

### LINKED ISSUE #2040: BUG : v3.14.0 throwing constant `globalThis.addEventListener is not a function` errors
**Describe the bug**
Since moving to v3.14.0 of @rudderstack/analytics-js, when running a local dev environment for Next.js, the console is filled with `TypeError: globalThis.addEventListener is not a function` errors pointing to the Rudderstack library.

**To Reproduce**

1. Create a new NextJS application `npx create-next-app`
2. Initialize Rudderstack in `layout.tsx`
3. Add a useEffect to `page.tsx` calling `rudderAnalytics.page` (make sure you add `use client` at the top of the page.tsx file since we're using useEffect)
4. Look in the console and you should see the errors that look like the image below

**Expected behavior**
There should be no errors related to Rudderstack in the console.

**Screenshots**

Image

**Additional Information (please complete the following information):**

- SDK installation type: NPM
- SDK version: 3.14.0
- Node version: v22.12.0
- NPM version: N/A (using `bun` )
- TypeScript version: v5.7.1
- Webpage URL: N/A
- Share the event payload: (if applicable)
- Integration that has the issue: (if applicable)
- Framework and version(e.g: Next.js, React, Vue): Next.js v 15.1.17
- Bundling toolset (e.g: Webpack, Rollup): Webpack

**Desktop (please complete the following information):**

- OS: macOS Sequoia 15.3
- Browser: Arc Browser v1.82.0

**Additional context**
I was seeing this in a repo with a lot of extra libraries and content, so I created a brand new project to see if I could reproduce, and I could. To confirm it wasn't just an issue with the `bun` runtime, I ran the new project with `npm`, `pnpm` and `bun` and all 3 exhibited the same issue.

saikumarrs:
## PR Description

I've fixed a couple of issues in the core module that would otherwise execute some logic before the SDK initialisation.

We create default instances of some resources so we should be careful in executing some code in their constructors. Otherwise, some code can get executed just by importing the package.

- Attaching error event handlers
- All storage engines default configuration

All the storage engine modules are refactored to create default instances which will be configured only when the SDK instance is created.
The default instances will be used throughout the code.
Now, there is no need for global singletons.

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-3003/avoid-attaching-error-listeners-by-default

## Cross Browser Tests

- [x] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

- **New Features**
  - Introduced dynamic configuration for analytics, allowing flexible key and endpoint setup.
  - Added a centralized initialization routine for global resources, enhancing error handling and storage services.

- **Refactor**
  - Streamlined the initialization process and configuration for various storage modules to improve consistency and reliability.
  - Simplified error handling setup to ensure smoother operation during runtime and testing.
  - Updated storage engine configuration to remove explicit enabling, relying on default behavior.
  - Enhanced the `ErrorHandler` class with a new initialization method for improved setup.
  - Modified storage classes to remove unnecessary options parameters, simplifying their constructors.

### DIFF
--- a/packages/analytics-js/src/app/RudderAnalytics.ts
+++ b/packages/analytics-js/src/app/RudderAnalytics.ts
@@ -38,6 +38,10 @@ import { defaultLogger } from '../services/Logger/Logger';
 import { PAGE_UNLOAD_ON_BEACON_DISABLED_WARNING } from '../constants/logMessages';
 import { state } from '../state';
 import { defaultErrorHandler } from '../services/ErrorHandler';
+import { defaultCookieStorage } from '../services/StoreManager/storages/CookieStorage';
+import { defaultLocalStorage } from '../services/StoreManager/storages/LocalStorage';
+import { defaultSessionStorage } from '../services/StoreManager/storages/sessionStorage';
+import { defaultInMemoryStorage } from '../services/StoreManager/storages/InMemoryStorage';
 
 // TODO: add analytics restart/reset mechanism
 
@@ -110,6 +114,12 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
   static initializeGlobalResources() {
     // We need to initialize the error handler first to catch any unhandled errors occurring in this module as well
     defaultErrorHandler.init();
+
+    // Initialize the storage engines with default options
+    defaultCookieStorage.configure();
+    defaultLocalStorage.configure();
+    defaultSessionStorage.configure();
+    defaultInMemoryStorage.configure();
   }
 
   /**
--- a/packages/analytics-js/src/app/RudderAnalytics.ts
+++ b/packages/analytics-js/src/app/RudderAnalytics.ts
@@ -168,7 +168,6 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
 
       setExposedGlobal(GLOBAL_PRELOAD_BUFFER, clone(preloadedEventsArray));
 
-      this.analyticsInstances[writeKey] = new Analytics();
       this.getAnalyticsInstance(writeKey)?.load(
         writeKey,
         dataPlaneUrl,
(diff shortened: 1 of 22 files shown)


packages/analytics-js/src/services/StoreManager/Store.ts
```diff
@@ -38,7 +38,7 @@ class Store implements IStore {
     this.name = config.name;
     this.isEncrypted = config.isEncrypted ?? false;
     this.validKeys = config.validKeys ?? {};
-    this.engine = engine ?? getStorageEngine(LOCAL_STORAGE);
+    this.engine = engine;
```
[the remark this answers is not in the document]
saikumarrs: Don't you think the type-safety is enough?

packages/analytics-js/src/app/RudderAnalytics.ts
```diff
@@ -168,7 +186,6 @@ class RudderAnalytics implements IRudderAnalytics<IAnalytics> {
 
       setExposedGlobal(GLOBAL_PRELOAD_BUFFER, clone(preloadedEventsArray));
 
-      this.analyticsInstances[writeKey] = new Analytics();
```
saikumarrs, line 171: This is not needed as `getAnalyticsInstance` call before will create a new instance if not already present.

packages/analytics-js/src/services/StoreManager/Store.ts
saikumarrs, line 41: This is an internal class and `engine` is a required parameter. So, having a default value does not make sense.

packages/analytics-js/src/services/StoreManager/StoreManager.ts
```diff
@@ -65,11 +65,10 @@ class StoreManager implements IStoreManager {
...
         sameDomainCookiesOnly: loadOptions.sameDomainCookiesOnly,
-        enabled: true,
       },
-      localStorageOptions: { enabled: true },
-      inMemoryStorageOptions: { enabled: true },
-      sessionStorageOptions: { enabled: true },
+      localStorageOptions: {},
+      inMemoryStorageOptions: {},
+      sessionStorageOptions: {},
     };
```
saikumarrs, line 72: These are already present in the default options of the respective module. It doesn't make sense to pass any values unless we're overriding them.

koladilip (approved):

PR #2057, opened 2025-02-21
feat: speed up rollback process

saikumarrs:
## PR Description

I've revamped the SDK version rollback workflow to copy artifacts directly from AWS S3 -> S3 to speed up the process.

Now, as we're not re-building the SDK artifacts at all, it should be really fast.

In case of the latest SDK, we're just copying the core SDK artifacts as it'll refer to the plugins and integrations from the versioned directories automatically.

For legacy SDK, all the artifacts are copied from the source.

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-2732/speed-up-the-sdk-rollback-process

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

- **Chores**
  - Updated the analytics integration to load from new, validated endpoints and credentials.
  - Enhanced configuration settings to improve consistency in how analytics components manage their versioning.

### DIFF
--- a/.github/workflows/rollback.yml
+++ b/.github/workflows/rollback.yml
@@ -9,7 +9,6 @@ on:
         type: choice
         options:
           - development
-          - staging
           - production
         default: 'development'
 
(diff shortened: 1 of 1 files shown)


saikumarrs:
> On GitHub, it will show the latest release right so will there be a separate effort to fix the issue and do a hotfix?

Ideally, we'll follow up with a new hotfix release to address the bugs right after the rollout. So, it may be fine.

However, I guess it's better mark the reverted version as the latest on GitHub and also deprecate the NPM package and replace the `latest` tag for a different version.

Yeah, let me create another ticket for this effort as they are not super critical for the rollback itself.

koladilip (dismissed):

ChryssaAliferi (dismissed):

PR #2088, opened 2025-03-05
feat: remove page loaded event

saikumarrs:
## PR Description

I've updated the page lifecycle auto-tracking feature to only send `Page Unloaded` event. This is to prevent duplicate events if users already have a default `page` event.

The page view ID and time spent on the current page parameters are both path of the `Page Unloaded` event. So, no information is truly lost.

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-2933/remove-page-loaded-option

## Cross Browser Tests

- [x] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

- **Refactor**
  - Streamlined analytics event tracking to focus exclusively on page unload events by removing previous page load tracking.
  - Removed the `LOADED` enum value from page lifecycle events.

- **Tests**
  - Expanded test coverage and improved state management to validate correct lifecycle event tracking and behavior under various conditions.
  - Introduced new tests for simulating page unload events and verifying event tracking based on the `useBeacon` option.

### DIFF
--- a/packages/analytics-js-common/src/types/LoadOptions.ts
+++ b/packages/analytics-js-common/src/types/LoadOptions.ts
@@ -109,7 +109,6 @@ export type PreConsentOptions = {
 };
 
 export enum PageLifecycleEvents {
-  LOADED = 'Page Loaded',
   UNLOADED = 'Page Unloaded',
 }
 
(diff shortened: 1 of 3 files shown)


koladilip (dismissed):
why are we removing

ChryssaAliferi (dismissed):
Great!! 🚀

saikumarrs:
> why are we removing

The product team believes that it can cause duplicate events if users are not careful. Users might already have a default `page` event. The `Page Loaded` event would just be a duplicate in that case.

koladilip (approved):

PR #2187, opened 2025-04-22
fix: recursively migrate persisted entries

saikumarrs:
## PR Description

I've updated the SDK v3 and v1.1 (legacy) packages to recursively decrypt cookies from both legacy and v3 (new) encryption formats until it is not encrypted anymore.

For v3 SDK, this happens as part of the cookie migration during the SDK initialization.

Without these fixes, a user visiting one more websites back and forth where either of the websites using v1.1 (a really old version) or v3 SDK packages, would end up in unencrypted values in the event payloads. 
A really old version of the v1.1 SDK cannot decrypt v3 cookies so it re-encrypts the already encrypted value. If this process repeats enough number of times, even the v3 SDK that can understand the v1.1 cookies will only decrypt once but the underlying value is still a encrypted value that is included in the events.

I also fixed the logic to only return the value from the state in the get APIs. Earlier, user ID and anonymous ID were returning values from the storage.

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-3225/handle-recursive-cookie-data-encryptions-js-sdk

## Cross Browser Tests

- [x] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

- **New Features**
  - Enhanced storage decryption to support recursive multi-layer decryption, ensuring stored data is fully decrypted regardless of legacy or current encryption methods.
  - Improved control over storage migration, allowing explicit selection of stores to migrate and ensuring migration occurs before reading stored entries.

- **Bug Fixes**
  - Refined anonymous ID retrieval to prioritize session state over storage, providing more consistent user identification.

- **Refactor**
  - Streamlined environment variable checks for configuration options in initialization scripts.
  - Updated internal logic for determining previous user ID in alias events.
  - Simplified Rollup configuration to watch additional source directories for changes.

- **Chores**
  - Removed obsolete warning messages and related test cases concerning legacy encrypted cookie data.
  - Expanded test coverage for alias event behavior and storage migration, including recursive decryption scenarios.
  - Added enhanced type annotations in user session manager tests to improve code clarity.

### DIFF
--- a/packages/analytics-js-common/src/v1.1/utils/storage/storage.js
+++ b/packages/analytics-js-common/src/v1.1/utils/storage/storage.js
@@ -30,9 +30,6 @@ const anonymousIdKeyMap = {
 const CORRUPTED_COOKIES_WARNING = key =>
   `Unable to retrieve the cookie data for ${key}. The data is dropped. This can potentially stem from using SDK v3 on other sites or web pages that can share cookies with this webpage. Please use the same SDK (v3) version everywhere as soon as possible.`;
 
-const BAD_COOKIES_WARNING = key =>
-  `The cookie data for ${key} seems to be corrupted where the underlying unencrypted data seems to be a cookie created by SDK v3. The decryption of the data was attempted again. This can potentially stem from using SDK v3 on other sites or web pages that can share cookies with this webpage. We recommend using the same SDK (v3) version everywhere.`;
-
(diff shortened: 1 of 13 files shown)


ChryssaAliferi (approved):
Looks good!

PR #2190, opened 2025-04-25
chore: use klaviyo push function to send events

maheshkutty:
## PR Description

Klaviyo recommend using `klaviyo` object instead of `_learnq` object for sending events

Resolves INT-3442

## Linear task (optional)

https://linear.app/rudderstack/issue/INT-3442/update-device-mode-to-use-klaviyopush-instead-of-learnqpush

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


yashasvibajpai (approved):

packages/analytics-js-integrations/src/integrations/Klaviyo/browser.js
```diff
@@ -93,7 +93,7 @@ class Klaviyo {
   }
 
   isLoaded() {
-    return !!(window._learnq && window._learnq.push !== Array.prototype.push);
+    return !!(window.klaviyo && window.klaviyo.push !== Array.prototype.push);
```
utsabc, line 96: Please share e2e Test report

packages/analytics-js-integrations/src/integrations/Klaviyo/browser.js
```diff
@@ -159,15 +159,15 @@ class Klaviyo {
...
+        window.klaviyo.push(['track', eventName, payload]);
       }
     } else {
       const propsPayload = properties;
       if (propsPayload.revenue) {
         propsPayload.$value = propsPayload.revenue;
         delete propsPayload.revenue;
       }
-      window._learnq.push(['track', event, propsPayload]);
+      window.klaviyo.push(['track', event, propsPayload]);
```
utsabc, line 170: What is the release strategy? How many customers will be affected by this change? What is the monitoring strategy in case something goes wrong.

Please keep a rollback strategy in place in case this goes wrong

utsabc (changes requested):

packages/analytics-js-integrations/src/integrations/Klaviyo/browser.js
maheshkutty: First things, I have tested this change on two customers. There is very little chance that things will go wrong
If it still happens, we can roll back this change by doing hotfix, or you can see the blocker here.
I identified two customers who are using Kalviyo in device mode:
* mood
* huckberry

We can monitor them after the release happens

packages/analytics-js-integrations/src/integrations/Klaviyo/browser.js
maheshkutty: Shared a video for mood.com
I have tested this thing using this Chrome tool. @saikumarrs shared this link
https://developer.chrome.com/docs/devtools/overrides

utsabc (approved):

PR #2295, opened 2025-06-17
feat: custom integration interface and types

MoumitaM:
## PR Description

- TypeScript interfaces and types for the custom device mode integrations feature
-  Moved analytics instance type from device-mode destination plugin to common package 

## Linear task (optional)

[Linear task link](https://linear.app/rudderstack/issue/SDK-3406/interface-and-types-definition)

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


packages/analytics-js/src/index.ts
```diff
@@ -14,9 +14,12 @@ export { type CookieSameSite } from '@rudderstack/analytics-js-common/types/Stor
 export { type ApiCallback, type ApiOptions } from '@rudderstack/analytics-js-common/types/EventApi';
 export { type ApiObject } from '@rudderstack/analytics-js-common/types/ApiObject';
 export { type IntegrationOpts } from '@rudderstack/analytics-js-common/types/Integration';
-export { type LogLevel } from '@rudderstack/analytics-js-common/types/Logger';
+export { type LogLevel, type RsaLogger } from '@rudderstack/analytics-js-common/types/Logger';
```
saikumarrs: Let's use this prefix everywhere.
```suggestion
export { type LogLevel, type RSALogger } from '@rudderstack/analytics-js-common/types/Logger';
```

packages/analytics-js-common/src/types/IRudderAnalytics.ts
```diff
@@ -197,3 +198,22 @@ export interface IRudderAnalytics<T = any> {
...
+  | 'getAnonymousId'
+  | 'getUserId'
+  | 'getUserTraits'
+  | 'getGroupId'
+  | 'getGroupTraits'
+  | 'getSessionId'
+> & {
+  loadIntegration: boolean;
+  logLevel: LogLevel;
+  loadOnlyIntegrations: IntegrationOpts;
```
saikumarrs, line 218: The properties are not needed for custom integrations but required for standard device mode integrations.

packages/analytics-js-common/src/types/CustomIntegration.ts
```diff
@@ -0,0 +1,85 @@
...
+   * @param event - The alias event payload to process
+   * @optional
+   */
+  alias?(analytics: RsaAnalyticsInstance, logger: RsaLogger, event: RsaEvent): void;
+}
+
+/**
+ * Type for the custom integration to be used in addCustomIntegration API
+ */
+export type RsaCustomIntegration = Pick<
```
saikumarrs: Have you check if this new type automatically picks the JSDoc information along with the properties? Otherwise, we have to do this differently.

packages/analytics-js-common/src/types/CustomIntegration.ts
```diff
@@ -0,0 +1,85 @@
...
+   */
+  name: string;
+
+  /**
+   * Initialize the integration
+   * @param analytics - The RudderStack analytics instance
+   * @param logger - The logger instance for this integration
+   * @optional
+   */
+  init?(analytics: RsaAnalyticsInstance, logger: RsaLogger): void;
```
saikumarrs: All the below properties from here should be arrow functions.

packages/analytics-js-common/src/types/ApplicationState.ts
```diff
@@ -186,6 +187,8 @@ export type ServerCookiesState = {
   dataServiceUrl: Signal<string | undefined>;
 };
 
+export type CustomIntegrationsState = Signal<CustomIntegration[]>;
```
saikumarrs: I doubt if we need a new state for customer integrations. The provided integration becomes the destination instance. 

I believe we can add them to the native integrations state itself but somehow bypass the loading step that we normally do for regular integrations

packages/analytics-js-common/src/types/ApplicationState.ts
saikumarrs, line 1: Please update the GitHub workflows to run them for `feature/sdk-3400-implementation-custom-device-mode-integrations` (parent branch) similarly to the `develop` branch.

saikumarrs (changes requested):

packages/analytics-js-common/src/types/CustomIntegration.ts
saikumarrs: `init` and rest of them should be arrow functions.

saikumarrs (dismissed):

packages/analytics-js-common/src/types/CustomIntegration.ts
MoumitaM: You cannot define an arrow function inside an interface; when the user implements this, they have to use an arrow function.

saikumarrs: They have to be arrow functions to support both CDN and NPM users. It goes without saying that we need to convert it into a type.

MoumitaM: Addressed

saikumarrs (approved):

PR #2411, opened 2025-07-29
fix: prevent null/undefined object conversion error in facebook pixel

maheshkutty:
## PR Description

## Linear task (optional)

Linear task link

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security


packages/analytics-js-integrations/src/integrations/FacebookPixel/utils.js
```diff
@@ -298,6 +315,9 @@ const getProductListViewedEventParams = properties => {
       quantity: 1,
     });
     contentType = 'product_group';
+  } else {
+    // Default to 'product' when no contentIds and no category
+    contentType = 'product';
```
krishna2020: this will change the existing behaviour - ideally if no contents/products are being sent, it's invalid event & shouldn't get considered as `product` content type

koladilip (approved):

krishna2020 (approved):

PR #2420, opened 2025-08-01
fix: add error handling to Heap integration to prevent SDK loading failures

koladilip:
## PR Description

This PR fixes the 'window.heap.push is not a function' error that was causing crashes in the Heap integration when the SDK failed to load properly.

### Problem
The Heap integration was throwing unhandled exceptions when:
- The Heap SDK script failed to load
- The SDK was blocked by ad blockers
- Network issues prevented SDK loading
- The SDK loaded but functions weren't available yet

### Solution
- Enhanced `isLoaded()` method to check for actual function availability (track, identify, addUserProperties)
- Added try-catch blocks around SDK calls in `track()` and `identify()` methods
- Added proper error logging when SDK is not available
- Implemented graceful degradation - events are logged as errors but don't break the application

### Changes
- `packages/analytics-js-integrations/src/integrations/Heap/browser.js`:
  - Improved `isLoaded()` to verify all required functions exist
  - Added error handling to `track()` method
  - Added error handling to `identify()` method
  - Added descriptive error messages for debugging

### Benefits
- Prevents application crashes from SDK loading failures
- Provides clear error messages for debugging
- Maintains compatibility with existing error reporting system
- Follows established error handling patterns in the codebase

## Linear task (optional)

N/A

## Cross Browser Tests

- [x] Chrome
- [x] Firefox
- [x] IE11

## Sanity Suite

## Security


koladilip:
@saikumarrs does it make sense to put integration login try catch as bugsnag won't be able catch it right

packages/analytics-js-integrations/src/integrations/Heap/browser.js
```diff
@@ -41,20 +46,38 @@ class Heap {
...
 
   identify(rudderElement) {
-    const { userId, context } = rudderElement.message;
-    if (userId) {
-      window.heap.identify(userId);
+    try {
+      if (!this.isLoaded()) {
+        logger.error('Heap SDK not loaded. Skipping identify call.');
+        return;
+      }
```
saikumarrs: If the integration did not report ready, no tracking APIs will be called. So, this check is redundant.

saikumarrs:
> @saikumarrs does it make sense to put integration login try catch as bugsnag won't be able catch it right

Yes, we won't be able to report these errors if they are handled in the integration. One option is to re-throw after some extra logging or clean up etc operations.

packages/analytics-js-integrations/src/integrations/Heap/browser.js
koladilip: Removed it

maheshkutty (approved):

PR #2441, opened 2025-08-18
feat: add default consent granted to Bing Ads integration

manish339k:
## PR Description

# Add Consent Management to Bing Ads Integration

## Summary
Implements Microsoft's consent compliance requirements by setting default consent state as "granted" when Bing Ads integration initializes

## Changes
- Added consent setting in `init`
- Updated test

## Consent Management
- **Default**: Consent is set to `'granted'` for `'ad_storage'` on initialization  Screenshot 2025-08-18 at 5 45 40 PM

- **Mid-session changes**: Use `window.bing{{tagId}}.push('consent', 'default', { 'ad_storage': 'granted' })` or `'denied'` Screenshot 2025-08-18 at 5 40 57 PM

## Testing
- ✅ Fixed test setup to handle consent calls
- ✅ Verified existing functionality preserved

## Impact
- No breaking changes
- Automatic consent compliance for customers
- Meets Microsoft's requirements

## Linear task (optional)

https://linear.app/rudderstack/issue/INT-3975/support-consent-parameters-update-for-bingads

## Cross Browser Tests

- [x] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

* **New Features**
  * Bing Ads integration now sends a default consent signal on initialization, granting ad storage to enable immediate operation of the SDK.

* **Tests**
  * Updated Bing Ads tracking tests to account for additional emitted events, adjusting expectations accordingly.

### DIFF
--- a/packages/analytics-js-integrations/src/integrations/BingAds/browser.js
+++ b/packages/analytics-js-integrations/src/integrations/BingAds/browser.js
@@ -32,6 +32,10 @@ class BingAds {
 
   init() {
     loadNativeSdk(this.uniqueId, this.tagID);
+    // Add consent as "granted" immediately when integration initializes
+    // If execution reaches here, user has consented or consent is not required
+    window[this.uniqueId] = window[this.uniqueId] || [];
+    window[this.uniqueId].push('consent', 'default', { ad_storage: 'granted' });
   }
 
   isLoaded() {
--- a/packages/analytics-js-integrations/src/integrations/BingAds/nativeSdkLoader.js
+++ b/packages/analytics-js-integrations/src/integrations/BingAds/nativeSdkLoader.js
@@ -1,12 +1,6 @@
 import { LOAD_ORIGIN } from '@rudderstack/analytics-js-legacy-utilities/constants';
 
 function loadNativeSdk(uniqueId, tagID) {
-  // Add consent as "granted" immediately when integration initializes
-  // If execution reaches here, user has consented or consent is not required
-  // UET script will process this consent and set the consent status accordingly
-  window[uniqueId] = window[uniqueId] || [];
-  window[uniqueId].push('consent', 'default', { ad_storage: 'granted' });
-
   ((w, d, t, r, u) => {
     let f;
     let n;


packages/analytics-js-integrations/src/integrations/BingAds/browser.js
```diff
@@ -35,9 +35,16 @@ class BingAds {
...
-      !!window.UET && !!window[this.uniqueId] && window[this.uniqueId].push !== Array.prototype.push
-    );
+    const isLoaded =
+      !!window.UET &&
+      !!window[this.uniqueId] &&
+      window[this.uniqueId].push !== Array.prototype.push;
+    if (isLoaded) {
+      // If integration loads successfully, user has consented or consent is not required
+      // In either case, it's safe to set consent as "granted"
+      window[this.uniqueId].push('consent', 'default', { ad_storage: 'granted' });
```
saikumarrs: I don't think you have to wait for third-party SDK to load successfully to queue the consent call.

The objective was that if the execution control reached this initialisation point in the integration, then we can assume that either user has explicitly consented to loaded or consent is not required to load this integration.

So, you can do this as part of `init` itself I guess (as long as it is allowed).

saikumarrs: CC @krishna2020

maheshkutty: Is there any documentation related to this? I see you've experimented with my calling this api

manish339k: I was first trying to do it in the init for the same reason but It was not resulting in expected result until bing ads sdk is loaded. Let me verify one more time.

saikumarrs: From their documentation, it seems like it is possible. Sure, please verify.

manish339k: https://help.ads.microsoft.com/#apex/ads/en/60119/1-500

manish339k: moved it to init.

packages/analytics-js-integrations/src/integrations/BingAds/nativeSdkLoader.js
```diff
@@ -1,6 +1,12 @@
 import { LOAD_ORIGIN } from '@rudderstack/analytics-js-legacy-utilities/constants';
 
 function loadNativeSdk(uniqueId, tagID) {
+  // Add consent as "granted" immediately when integration initializes
+  // If execution reaches here, user has consented or consent is not required
+  // UET script will process this consent and set the consent status accordingly
+  window[uniqueId] = window[uniqueId] || [];
```
maheshkutty: This code block should be after we call `loadNativeSdk` function call?
cc: @saikumarrs

saikumarrs: Yes. Wrong placement @manish339k.

manish339k: Not necessarily. We are putting consent in the queue before loading the script. I think it's better because consent is always processed first in this way.

saikumarrs: `loadNativeSdk` function is meant for placing only the raw initialisation snippet(s) from the third-party tools.

This kind of initialisation logic should go into the `init` method right after the `loadNativeSdk` call. 
Even in this way, `consent` is still the first item to be processed in the queue.

manish339k: got it.

manish339k: resolved it.

packages/analytics-js-integrations/src/integrations/BingAds/browser.js
```diff
@@ -32,6 +32,10 @@ class BingAds {
 
   init() {
     loadNativeSdk(this.uniqueId, this.tagID);
+    // Add consent as "granted" immediately when integration initializes
+    // If execution reaches here, user has consented or consent is not required
+    window[this.uniqueId] = window[this.uniqueId] || [];
```
saikumarrs, line 37: Ideally, this line is not required as the initialisation snippet is already doing it.

saikumarrs (approved):

packages/analytics-js-integrations/src/integrations/BingAds/browser.js
manish339k: Yeah, but suppose for some unknown reason it does not exist, then I think in that case it is good to have. right?

saikumarrs: If initialisation snippet failed, the integration is not functional anyway. It does not cause any harm, so feel free to keep it.

PR #2585, opened 2025-11-12
feat: add support of Platform Specific App Keys for braze

ItsSudip:
## PR Description

We are adding support of platform specific appKey support for braze destination. It will be a backward compatible change through `usePlatformSpecificApiKeys` key in config. If the customer enabled `usePlatformSpecificApiKeys` key then only we will look for then platform specific appKey i.e. `webAppKey` for Javascript SDK.

## Linear task (optional)

https://linear.app/rudderstack/issue/INT-4362/js-add-support-for-the-braze-multi-app-key-support

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

* **New Features**
  * Added a configuration flag to enable platform-specific app keys for the Braze integration, allowing a web-specific app key to override the default when enabled.
* **Bug Fixes**
  * When a provided platform-specific key is invalid, a warning is emitted and the integration falls back to the default key.
* **Tests**
  * Added tests covering the new flag behavior and the invalid-key warning/fallback.

### DIFF
--- a/packages/analytics-js-integrations/src/integrations/Braze/browser.js
+++ b/packages/analytics-js-integrations/src/integrations/Braze/browser.js
@@ -21,14 +21,21 @@ class Braze {
     }
     this.analytics = analytics;
     this.appKey = config.appKey;
-    this.usePlatformSpecificAppKeys = config.usePlatformSpecificApiKeys || false;
+    this.usePlatformSpecificAppKeys = config.usePlatformSpecificAppKeys === true;
     this.trackAnonymousUser = config.trackAnonymousUser;
     this.enableBrazeLogging = config.enableBrazeLogging || false;
     this.allowUserSuppliedJavascript = config.allowUserSuppliedJavascript || false;
     this.enablePushNotification = config.enablePushNotification || false;
     if (!config.appKey) this.appKey = '';
-    if (this.usePlatformSpecificApiKeys && config.webAppKey) {
-      this.appKey = config.webAppKey;
+    if (this.usePlatformSpecificAppKeys) {
+      if (config.webAppKey && typeof config.webAppKey === 'string') {
+        this.appKey = config.webAppKey;
+      } else {
+        logger.warn(
+          'Configured to use platform specific app key but the web app key is not valid:',
+          config.webAppKey,
+        );
+      }
     }
     this.endPoint = '';
     this.isHybridModeEnabled = config.connectionMode === 'hybrid';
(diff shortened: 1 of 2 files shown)


packages/analytics-js-integrations/src/integrations/Braze/browser.js
```diff
@@ -21,11 +21,15 @@ class Braze {
     }
     this.analytics = analytics;
     this.appKey = config.appKey;
+    this.usePlatformSpecificAppKeys = config.usePlatformSpecificApiKeys || false;
```
saikumarrs: @1abhishekpandey it should be technically named `usePlatformSpecificAppKeys`, right? API key is the term used for Braze cloud connection.

packages/analytics-js-integrations/src/integrations/Braze/browser.js
```diff
@@ -21,11 +21,15 @@ class Braze {
...
     this.trackAnonymousUser = config.trackAnonymousUser;
     this.enableBrazeLogging = config.enableBrazeLogging || false;
     this.allowUserSuppliedJavascript = config.allowUserSuppliedJavascript || false;
     this.enablePushNotification = config.enablePushNotification || false;
     if (!config.appKey) this.appKey = '';
+    if (this.usePlatformSpecificApiKeys && config.webAppKey) {
```
saikumarrs: @ItsSudip, let's be strict with the validations and also not silently fail.

```suggestion
    if (this.usePlatformSpecificApiKeys) {
      if (config.webAppKey && typeof config.webAppKey === 'string') {
        this.appKey = config.webAppKey;
      } else {
        logger.warn("Configured to use platform specific app key but the web app key is not valid:", config.webAppKey);
      }
    }
```

packages/analytics-js-integrations/src/integrations/Braze/browser.js
saikumarrs: This would be resilient.

```suggestion
    this.usePlatformSpecificAppKeys = config.usePlatformSpecificApiKeys === true;
```

packages/analytics-js-integrations/src/integrations/Braze/browser.js
saikumarrs, line 1: @ItsSudip, let's add unit tests for the changes.

saikumarrs (changes requested):

packages/analytics-js-integrations/src/integrations/Braze/browser.js
1abhishekpandey: I believe you're right, we can improve it to have proper naming, I'll improve mobile integrations to use this key.

@ItsSudip Let's make this change.

```suggestion
    this.usePlatformSpecificAppKeys = config.usePlatformSpecificAppKeys || false;
```

ItsSudip: Done

packages/analytics-js-integrations/src/integrations/Braze/browser.js
ItsSudip: Added test cases

PR #2623, opened 2025-12-01
feat: allow overloaded identify method in query string api

saikumarrs:
## PR Description

I've enhanced the query string API in the SDK to allow setting traits (via identify) call for anonymous users.

The identify method supports two overloads,
```
rudderanalytics.identify(userId, traits);

// Updates traits for existing user (identified or anonymous)
rudderanalytics.identify(traits);
```

Before this change, the SDK did not fire any `identify` event when the page URL was something like below,
```
https://<page_url>?ajs_trait_example_key_1=example_value_1&ajs_trait_example_key_2=example_value_2
```

Now, it does.

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-4239/improve-query-string-api

## Cross Browser Tests

- [x] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

### DIFF
--- a/packages/analytics-js/__tests__/components/preloadBuffer/index.test.ts
+++ b/packages/analytics-js/__tests__/components/preloadBuffer/index.test.ts
@@ -51,6 +51,30 @@ describe('Preload Buffer', () => {
     ]);
   });
 
+  it('should retrieve identify event with only traits (no userId)', () => {
+    const testUrlParams = `${originalWindowLocation}?ajs_trait_email=test@example.com&ajs_trait_name=John`;
+
+    window.history.pushState({}, '', testUrlParams);
+
+    const argumentsArray: PreloadedEventCall[] = [];
+    retrieveEventsFromQueryString(argumentsArray);
+
+    expect(argumentsArray).toStrictEqual([
+      ['identify', { email: 'test@example.com', name: 'John' }],
+    ]);
+  });
+
+  it('should not add identify event when no userId or traits are present', () => {
+    const testUrlParams = `${originalWindowLocation}?ajs_event=dummyName&ajs_prop_dummy=true`;
+
+    window.history.pushState({}, '', testUrlParams);
+
+    const argumentsArray: PreloadedEventCall[] = [];
+    retrieveEventsFromQueryString(argumentsArray);
+
+    expect(argumentsArray).toStrictEqual([['track', 'dummyName', { dummy: 'true' }]]);
+  });
+
   it('should retrieve the load event if any', () => {
     const argumentsArray: PreloadedEventCall[] = [
       ['track'],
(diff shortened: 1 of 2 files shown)


packages/analytics-js/__tests__/components/preloadBuffer/index.test.ts
```diff
@@ -51,6 +51,30 @@ describe('Preload Buffer', () => {
...
+    const argumentsArray: PreloadedEventCall[] = [];
+    retrieveEventsFromQueryString(argumentsArray);
+
+    expect(argumentsArray).toStrictEqual([
+      ['identify', { email: 'test@example.com', name: 'John' }],
+    ]);
+  });
+
+  it('should not add identify event when no userId or traits are present', () => {
+    const testUrlParams = `${originalWindowLocation}?ajs_event=dummyName&ajs_prop_dummy=true`;
```
shekhar-rudder, line 68: What is the difference between the `testUrlParams` in both the tests? Are only specific params considered as traits (like email or name) ?

saikumarrs: `ajs_trait_<trait_name>` are considered as user traits.
`ajs_prop_<prop_name>` are considered as properties for track event.

More here,
https://www.rudderstack.com/docs/sources/event-streams/sdks/rudderstack-javascript-sdk/supported-api/#query-string-api

packages/analytics-js/src/components/preloadBuffer/index.ts
```diff
@@ -62,13 +63,14 @@ const retrieveEventsFromQueryString = (argumentsArray: PreloadedEventCall[] = []
...
-  if (queryObject.get(QUERY_PARAM_USER_ID_KEY)) {
-    argumentsArray.unshift([
-      'identify',
-      queryObject.get(QUERY_PARAM_USER_ID_KEY),
-      getEventDataFromQueryString(queryObject, eventArgumentToQueryParamMap.trait),
-    ]);
+  // Send identify event
+  const userId = queryObject.get(QUERY_PARAM_USER_ID_KEY);
+  const userTraits = getEventDataFromQueryString(queryObject, eventArgumentToQueryParamMap.trait);
+  if (userId || isNonEmptyObject(userTraits)) {
```
shekhar-rudder, line 69: Do we need to handle the scenario where `userId` is present but not `userTraits`?

saikumarrs: No. A user can be identified without any traits.

See this section for all the possible invocations of the API,
https://www.rudderstack.com/docs/sources/event-streams/sdks/rudderstack-javascript-sdk/supported-api/#identify

shekhar-rudder (approved):

PR #2636, opened 2025-12-04
fix: prefer state over storage values when server-side cookie requests are in progress

saikumarrs:
## PR Description

Fixes race conditions when reading user session data during in-flight server-side cookie requests. The SDK now prefers in-memory state over storage when async cookie operations are pending, preventing stale data reads.

## Problem

When `useServerSideCookies: true`:

1. User calls `setUserId('new-user')` → state updates immediately
2. Server-side cookie request queued (10ms debounce)
3. **Race condition**: `getUserId()` reads from storage (old value) while state has new value ❌

Additional issues:
- Multiple tabs updating same cookie concurrently
- `refreshSession()` reading stale values during async operations

## Solution

**The Fix**: Track when cookie requests are in-progress and read from the right source:
- **During async requests** → read from state (has latest value)
- **After requests complete** → read from storage (synchronized source of truth)

### Implementation

1. **Added tracking mechanism** - Flag to know when server-side cookie requests are pending
2. **Centralized getter logic** - New `getUserSessionValue()` method that checks the flag and picks the right data source
3. **Updated all getters** - All 8 session data getters (`getUserId()`, `getUserTraits()`, etc.) now use this centralized logic
4. **Smarter error handling** - Don't log errors when other browser tabs update cookies concurrently

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-4207/fix-race-condition-with-session-data-when-server-side-cookies-are-used

## Cross Browser Tests

- [x] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

## Release Notes

* **Improvements**
  * Enhanced session state consistency during server-side cookie synchronization for more reliable data handling in concurrent operations.

* **Tests**
  * Expanded test coverage for session management across various server-side cookie request scenarios.

* **Chores**
  * Updated configuration parameters.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
+++ b/packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
@@ -690,11 +690,27 @@ class UserSessionManager implements IUserSessionManager {
   }
 
   /**
+   * Fetches the value for a session key. Preferably from storage, if the server-side
+   * cookies request is not in progress. Otherwise, from the state.
+   * @param sessionKey - The session key to fetch the value for
+   * @returns - The value for the session key
+   */
+  getUserSessionValue<T>(sessionKey: UserSessionKey): Nullable<T> {
+    // If the server-side cookies request is in progress, fetch the value from the state.
+    if (this.serverSideCookiesRequestInProgress[sessionKey]) {
+      return state.session[sessionKey].value as Nullable<T>;
+    }
+
+    // Otherwise, fetch the value from storage.
+    return this.getEntryValue(sessionKey);
+  }
+
+  /**
    * Fetches User Id
    * @returns
    */
   getUserId(): Nullable<string> {
-    return this.getEntryValue('userId');
+    return this.getUserSessionValue<string>('userId');
   }
 
   /**
@@ -702,7 +718,7 @@ class UserSessionManager implements IUserSessionManager {
    * @returns
    */
   getUserTraits(): Nullable<ApiObject> {
-    return this.getEntryValue('userTraits');
+    return this.getUserSessionValue<ApiObject>('userTraits');
   }
 
   /**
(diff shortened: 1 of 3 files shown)


saikumarrs:
@coderabbitai review

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
```diff
@@ -91,6 +91,10 @@ class UserSessionManager implements IUserSessionManager {
   httpClient: IHttpClient;
   logger: ILogger;
   serverSideCookieDebounceFuncs: Record<UserSessionKey, number>;
+  /**
+   * Tracks whether the setting the cookies action has been queued or not 1Code has comments. Press enter to view.
```
achettyiitr: What's 1code?

saikumarrs: Addressed.

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
```diff
@@ -449,6 +454,20 @@ class UserSessionManager implements IUserSessionManager {
       );
     });
 
+    const clearInProgressFlags = () => {
+      sessionKeys.forEach((sessionKey: UserSessionKey) => {
```
achettyiitr, line 458: Is it possible that sessionKeys got updated as well and UserSessionKey is not present. so serverSideCookiesRequestInProgress will still contain stale values?

achettyiitr: Can we not reinitialise `this.serverSideCookiesRequestInProgress`?

achettyiitr (approved):

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
saikumarrs: `sessionKeys` won't get updated as it's reference is locally saved in the function.

As with every request, only a subset of cookies (in fact, it's always 1 - we haven't optimised it yet) are update, we shouldn't reinitialise `this.serverSideCookiesRequestInProgress`.

PR #2956, opened 2026-04-20
fix(analytics-js-service-worker): bump axios to fix vulnerabilities

SKannaniOS:
## PR Description

Bumps `axios` from `1.13.6` to `1.15.1` to address two critical-severity security advisories (GHSA-fvcv-3m26-pcqx and GHSA-3p68-rc4w-qgx5). The version is updated in both the root `package.json` (monorepo hoisted dependency) and `packages/analytics-js-service-worker/package.json` (the only package that directly consumes axios). Axios is the HTTP client used in the service worker SDK for sending events to the data plane. The upgrade also pulls in `proxy-from-env` v2.1.0 as a transitive dependency update.

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-4758/bump-axios-and-publish-a-new-js-sdk-version

## Cross Browser Tests

This is a server-side/service worker package with no browser UI. The HTTP transport layer is covered by unit tests. Cross-browser testing is not applicable for this change.

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

Addresses:
- GHSA-fvcv-3m26-pcqx — Axios Unrestricted Cloud Metadata Exfiltration via Header Injection Chain
- GHSA-3p68-rc4w-qgx5 — Axios NO_PROXY Hostname Normalization Bypass leading to SSRF

## Summary by CodeRabbit

* **Chores**
  * Updated build configuration size limits for Service Worker artifacts.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "@segment/top-domain": "3.0.1",
         "@vespaiach/axios-fetch-adapter": "0.3.1",
         "assert": "2.1.0",
-        "axios": "1.13.6",
+        "axios": "1.15.1",
         "axios-retry": "4.5.0",
         "component-each": "0.2.6",
         "component-emitter": "2.0.0",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -2408,6 +2408,18 @@
         }
       }
     },
+    "node_modules/@commitlint/read/node_modules/conventional-commits-filter": {
```
saikumarrs, line 2411: @SKannaniOS this change doesn't make sense. Let's revert it.

SKannaniOS: The `node_modules/@commitlint/read/node_modules/conventional-commits-filter` entry was not introduced intentionally — it appeared automatically when `npm install` was run after bumping axios.

Root cause: `@commitlint/read `already required `conventional-commits-filter@^5.0.0` in the existing develop lock file, but only 4.0.0 was resolved at the root. This was a pre-existing inconsistency. Running `npm install` caused npm to properly resolve it by adding the nested 5.0.0 entry.

The entry is marked `dev`, `optional`, and `peer` — no production impact. Removing it would restore the previous inconsistent state.

Do you really want to remove that entry?

saikumarrs (dismissed):

saikumarrs:
@SKannaniOS let's bump the size limits of the affected package to resolve the failing check.

saikumarrs (approved):

PR #3128, opened 2026-07-29
feat: sdk-5057 add custom context validation foundation

dciccale:
## Summary

- add a reusable custom-context update preparation boundary for the JS SDK
- require a plain top-level object and filter reserved root context keys with value-free warnings
- extract recursive `null` and explicit `undefined` object-property deletion paths without creating marker-only branches
- reuse existing SDK sanitation for supported inputs, including `BigInt` and circular-reference handling
- validate retained values atomically, including finite numbers, valid `Date` values, plain objects, and arrays
- reject unsupported retained values and `null`/`undefined` array entries without partially applying an update
- defensively clone retained context values and deletion paths

## Behavior

This PR establishes the shared validation and preparation layer used by global custom-context writes. It does not add public runtime methods, instance state, load-time seeding, or event enrichment.

Reserved-key filtering is shallow and protects the root of event context while allowing the same names below customer-defined keys. Invalid input produces value-free warnings and no prepared update. Deletion markers are represented as arrays of path segments so keys containing dots remain unambiguous.

Arrays are validated and preserved for the SDK's existing `mergeDeepRight` behavior. Deletion markers inside arrays remain invalid. Valid `Date` values are retained and defensively cloned in accordance with the approved design.

## Validation

- focused custom-context utility suite: 24 tests passed
- `@rudderstack/analytics-js` lint: 0 errors; 58 existing repository warnings
- Prettier check passed for all changed files
- `git diff --check` passed

## Design

- [HLD](https://github.com/rudderlabs/rudder-specs/blob/master/sdk-architecture/js-custom-context/specification.md)
- [LLD PR](https://github.com/rudderlabs/rudder-specs/pull/106)
- [SDK-5057](https://linear.app/rudderstack/issue/SDK-5057/implement-reserved-key-filtering-and-warnings-for-global-custom)

## Summary by CodeRabbit

* **New Features**
  * Added validation and sanitization for custom context updates, including nested objects, arrays, dates, and deletion markers.
  * Filters reserved and unsafe keys with clear warnings.
  * Exposed custom context utilities and TypeScript types for integration.
* **Bug Fixes**
  * Improved handling of invalid values, circular references, shared objects, and prototype-pollution attempts.
  * Preserves immutability and normalizes cross-environment objects and dates.
* **Tests**
  * Added comprehensive coverage for sanitization, pruning, cloning, warnings, and browser cross-realm behavior.


packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,159 @@
...
+        const activeObjectIndex = activeObjects.indexOf(childValue);
+        if (activeObjectIndex >= 0) {
+          retainedValue[key] = activeRetainedValues[activeObjectIndex];
+          return;
+        }
+
+        const retainedChildValue = visitObject(childValue, childPath);
+        if (Object.keys(retainedChildValue).length > 0) {
+          retainedValue[key] = retainedChildValue;
+        }
```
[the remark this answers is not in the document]
dciccale: Fixed in cc12bbf8. Explicitly supplied empty objects are now retained, while branches containing only deletion markers are still pruned. Added a regression covering both cases.

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,159 @@
...
+type UnknownContext = Record<string, unknown>;
+
+const isPlainObject = (value: unknown): value is UnknownContext => {
+  if (value === null || typeof value !== 'object') {
+    return false;
+  }
+
+  const prototype = Object.getPrototypeOf(value);
+  return prototype === Object.prototype || prototype === null;
+};
```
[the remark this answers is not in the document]
dciccale: Fixed in cc12bbf8. The check now uses the SDK's shared object-literal classification and a realm-neutral prototype-chain check, so iframe-created plain objects are accepted while custom-prototype/class-like objects remain outside the plain-object contract. Added regressions for both boundaries.

packages/analytics-js/__tests__/components/customContext/utilities.test.ts
```diff
@@ -0,0 +1,250 @@
...
+
+    try {
+      expect(prepareCustomContextUpdate(crossRealmContext, logger)).toEqual({
+        context: { region: 'EU' },
+        deletionPaths: [],
+      });
+    } finally {
+      iframe.remove();
+    }
+  });
```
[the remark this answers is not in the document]
dciccale: Fixed in 241f2933. The iframe-based cross-realm regression now lives in a dedicated browser test suite, while the shared utility suite remains DOM-independent. The assertions and iframe cleanup are preserved.

dciccale:
CodeRabbit follow-up addressed in 241f2933: isolated the iframe regression in a browser-specific suite and strengthened the deletion-path independence assertion. I kept the dedicated prepared-context type because, unlike the shared ApiObject type, prepared snapshots must exclude null/undefined deletion markers and nullish array members per the finalized custom-context contract.

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,185 @@
...
+  if (!isValidCustomContextValue(sanitizedContext)) {
+    logger.warn(INVALID_CUSTOM_CONTEXT_WARNING(CUSTOM_CONTEXT));
+    return undefined;
+  }
+
+  return {
+    context: clone(sanitizedContext as CustomContext),
+    deletionPaths: inspectedUpdate.deletionPaths.map(path => [...path]),
+  };
+};
```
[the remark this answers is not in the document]
dciccale: Fixed in d9776c89. Custom-context preparation now rejects `__proto__`, `constructor`, and `prototype` keys recursively before accepted values are copied or sanitized, while continuing to skip filtered reserved-root values. Added regressions for root, nested, and array-contained keys and verified that object prototypes remain untouched.

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,185 @@
...
+    return true;
+  }
+
+  const constructor = Object.prototype.hasOwnProperty.call(prototype, 'constructor')
+    ? prototype.constructor
+    : undefined;
+
+  return (
+    typeof constructor === 'function' &&
+    Function.prototype.toString.call(constructor) === objectConstructorSource
```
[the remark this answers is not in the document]
dciccale: Fixed in ef9921ec. The realm-neutral plain-object check now also requires `constructor.prototype === prototype`, and a regression rejects a custom prototype that spoofs the native `Object` constructor. The cross-realm browser tests continue to pass.

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  if (isValidDate(value)) {
+    return true;
+  }
+
+  if (Array.isArray(value)) {
+    return [...value].every(
+      item => item !== null && item !== undefined && isValidCustomContextValue(item),
+    );
```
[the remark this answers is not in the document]
dciccale: Fixed in ef9921ec. Array validation now walks every numeric index and requires an own entry, so sparse arrays and custom iterators cannot bypass the retained-value contract. Added a custom-iterator regression alongside the existing sparse-array coverage.

packages/analytics-js/src/constants/logMessages.ts
```diff
@@ -158,6 +158,9 @@ const RESERVED_KEYWORD_WARNING = (
 const INVALID_CONTEXT_OBJECT_WARNING = (logContext: string): string =>
   `${logContext}${LOG_CONTEXT_SEPARATOR}Please make sure that the "context" property in the event API's "options" argument is a valid object literal with key-value pairs.`;
 
+const INVALID_CUSTOM_CONTEXT_WARNING = (logContext: string): string =>
+  `${logContext}${LOG_CONTEXT_SEPARATOR}The custom context update is invalid. Use a plain object containing only supported context values.`;
```
saikumarrs: What do we mean by "supported context values" here? Maybe we should clarify the "invalid" values that are not supported.

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,226 @@
...
+
+const CUSTOM_CONTEXT = 'CustomContext';
+const CUSTOM_CONTEXT_PARENT_KEY_PATH = 'custom context';
+const PROTOTYPE_POLLUTION_KEYS = new Set(['__proto__', 'constructor', 'prototype']);
+
+type UnknownContext = Record<string, unknown>;
+
+const objectConstructorSource = Function.prototype.toString.call(Object);
+
+const isPlainObject = (value: unknown): value is UnknownContext => {
```
saikumarrs, line 17: Should we instead update the `isObjectLiteralAndNotNull` implementation with these exhaustive checks?

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,226 @@
...
+  if (visitedObjects.has(traversableValue)) {
+    return false;
+  }
+
+  visitedObjects.add(traversableValue);
+
+  return Object.keys(traversableValue).some(key => {
+    if (isRoot && CONTEXT_RESERVED_ELEMENTS.includes(key)) {
+      return false;
+    }
```
saikumarrs: This check seems redundant. Can you please clarify the intent?

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,226 @@
...
+const filterReservedCustomContextKeys = (
+  context: UnknownContext,
+  logger: ILogger,
+): UnknownContext => {
+  const retainedContext: UnknownContext = {};
+
+  Object.keys(context).forEach(key => {
+    if (CONTEXT_RESERVED_ELEMENTS.includes(key)) {
+      logger.warn(
+        RESERVED_KEYWORD_WARNING(
```
saikumarrs: This warning is not the right fit for this use case. Please create a new one specific to custom context.

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,226 @@
...
+    return undefined;
+  }
+
+  if (containsPrototypePollutionKey(input, true)) {
+    logger.warn(INVALID_CUSTOM_CONTEXT_WARNING(CUSTOM_CONTEXT));
+    return undefined;
+  }
+
+  const acceptedInput = filterReservedCustomContextKeys(input, logger);
+  const inspectedUpdate = inspectDeletionMarkers(acceptedInput, input);
```
saikumarrs: Do we really need to inspect the deletion markers? Upon deep merge, they're automatically deleted.

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,226 @@
...
+  const sanitizedContext = getSanitizedValue(inspectedUpdate.context, logger);
+
+  if (!isValidCustomContextValue(sanitizedContext)) {
```
saikumarrs: I don't think we should be concerned of these invalid values as they don't have any impact on the processing.
We've already tackled prototype pollution, circular references etc. above which should be sufficient.

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,226 @@
...
+  CustomContextDeletionPath,
+  CustomContextValue,
+  PreparedCustomContextUpdate,
+} from './types';
+
+const CUSTOM_CONTEXT = 'CustomContext';
+const CUSTOM_CONTEXT_PARENT_KEY_PATH = 'custom context';
+const PROTOTYPE_POLLUTION_KEYS = new Set(['__proto__', 'constructor', 'prototype']);
+
+type UnknownContext = Record<string, unknown>;
```
saikumarrs: Nipick: Let's move this also to the types file.

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,226 @@
...
+
+const objectConstructorSource = Function.prototype.toString.call(Object);
```
saikumarrs: This can be scoped to the `isPlainObject` check. No other code is referencing it.

packages/analytics-js/src/components/customContext/utilities.ts
saikumarrs: We can use the existing utility function to delete the undefined values at the end.

packages/analytics-js/__tests__/components/customContext/utilities.test.ts
```diff
@@ -0,0 +1,269 @@
+import type { ILogger } from '@rudderstack/analytics-js-common/types/Logger';
+import { CONTEXT_RESERVED_ELEMENTS } from '../../../src/components/eventManager/constants';
+import { prepareCustomContextUpdate } from '../../../src/components/customContext';
+
+class MockLogger implements ILogger {
```
saikumarrs: Mock logger already exists here, `packages/analytics-js-common/__mocks__/Logger.ts`

saikumarrs (changes requested):

packages/analytics-js/src/constants/logMessages.ts
dciccale: That’s fair. One immediate option is to make the warning explicit:

```text
The custom context update is invalid. Use a plain object. Functions, symbols, non-finite numbers, invalid Dates, non-plain objects, sparse arrays, and null or undefined array items are not supported.
```

The tradeoff is that the full list adds bytes to the SDK bundle. Do we have, or would we want, a shortened documentation URL such as `https://rddr.co/4Hdu3` pointing directly to the valid-value contract? The warning could then remain compact, following the error-link pattern used by libraries such as React:

```text
Invalid custom context. See https://rddr.co/4Hdu3 for supported values.
```

The short-link option requires documentation and redirect work, so we could keep the explicit list for convenience now and defer that improvement. Do you have a preference?

packages/analytics-js/src/components/customContext/utilities.ts
dciccale: That’s a valid option. The caveat is that `isObjectLiteralAndNotNull` is shared across event parsing, JSON sanitation, persistence, configuration, and retry code, and its existing test explicitly accepts class instances; adding these checks there would tighten the behavior for every caller. We could try the shared change and validate for side effects, but I’d prefer to keep the stricter check isolated here so this PR’s behavioral impact remains scoped to custom context. Would you be okay with keeping it local?

packages/analytics-js/src/components/customContext/utilities.ts
dciccale: You’re right. This made me reconsider the responsibility boundary. The new flow will safely filter reserved root keys first, and then an agnostic `containsPrototypePollutionKey` will scan only the retained input: `{ screen: unsafe, plan: 'pro' }` becomes `{ plan: 'pro' }` before scanning, while `{ plan: { constructor: {} } }` remains and is rejected. This keeps the scanner focused on one responsibility without changing the custom-context contract.

packages/analytics-js/src/components/customContext/utilities.ts
dciccale: Added a custom-context-specific warning that names the reserved top-level key and makes it clear that the key was ignored.

packages/analytics-js/src/components/customContext/utilities.ts
dciccale: I checked this against the agreed custom-context contract. The LLD explicitly uses deletion paths and prunes marker-only branches so deleting a missing path remains a no-op. Merging first and then calling `removeUndefinedAndNullValues` handles existing paths, but `{ missing: { nested: null } }` would leave `{ missing: {} }` when `missing` did not previously exist, and it loses the distinction from an explicitly supplied empty object. Are you suggesting we change that contract, or would you be okay keeping the inspector for those semantics?

packages/analytics-js/src/components/customContext/utilities.ts
dciccale: I checked this against the agreed value contract. This validation ensures that values stored and returned by `getCustomContext()` survive event serialization predictably. Without it, functions and symbols disappear, `Map` and `Set` become empty objects, and non-finite numbers, invalid dates, or sparse array entries become `null`. The HLD and LLD currently require these updates to be rejected atomically, so I would prefer to keep the validation. Are you proposing that we broaden that contract?

packages/analytics-js/src/components/customContext/utilities.ts
dciccale: Moved `UnknownContext` into the custom-context types file alongside the related internal types.

packages/analytics-js/src/components/customContext/utilities.ts
dciccale: `isPlainObject` runs repeatedly while inspecting and validating nested objects, so I kept the native constructor source at module scope to compute it once per module load. I renamed it to `OBJECT_CONSTRUCTOR_SOURCE` to make the cached-constant intent clearer.

packages/analytics-js/__tests__/components/customContext/utilities.test.ts
dciccale: Reused the shared `Logger` mock in both custom-context test suites and removed the local duplicates.

packages/analytics-js/src/constants/logMessages.ts
```diff
@@ -155,9 +155,15 @@ const RESERVED_KEYWORD_WARNING = (
 ): string =>
   `${context}${LOG_CONTEXT_SEPARATOR}The "${property}" property defined under "${parentKeyPath}" is a reserved keyword. Please choose a different property name to avoid conflicts with reserved keywords (${reservedElements}).`;
 
+const RESERVED_CUSTOM_CONTEXT_KEY_WARNING = (logContext: string, property: string): string =>
+  `${logContext}${LOG_CONTEXT_SEPARATOR}The top-level custom context property "${property}" is reserved and was ignored.`;
```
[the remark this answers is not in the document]
dciccale: `property` reaches this helper only after an exact match against the static `CONTEXT_RESERVED_ELEMENTS` list (`library`, `consentManagement`, `userAgent`, `ua-ch`, and `screen`). A key containing quotes or control characters cannot reach this warning through `filterReservedCustomContextKeys`, so the reported log-forging path is not reachable. I’d keep the current formatting.

packages/analytics-js/src/components/customContext/utilities.ts
```diff
@@ -0,0 +1,219 @@
...
+  if (!isValidCustomContextValue(sanitizedContext)) {
+    logger.warn(INVALID_CUSTOM_CONTEXT_WARNING(CUSTOM_CONTEXT));
+    return undefined;
+  }
+
+  return {
+    context: clone(sanitizedContext as CustomContext),
+    deletionPaths: inspectedUpdate.deletionPaths.map(path => [...path]),
+  };
+};
```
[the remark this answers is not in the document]
dciccale: `prepareCustomContextUpdate` now fails closed around the complete preparation pipeline. Throwing accessors or Proxy traps produce the existing invalid-context warning and no prepared update, with regressions covering both cases.

packages/analytics-js/src/components/customContext/utilities.ts
saikumarrs: Let's create a follow up backlog ticket for the refactoring. Resolving the comment.

packages/analytics-js/src/components/customContext/utilities.ts
saikumarrs: I believe this is an edge case and since the data is user supplied, the side-effects should be expected if they are not careful.

To simplify our implementation, let's deviate from the contract.

packages/analytics-js/src/components/customContext/utilities.ts
saikumarrs: As this is user supplied data will have to be included in the events, serialised and sent over the network, they can't use any values that don't survive it.

I think this is a good trade off. It is justified for prototype pollution and others as it can mess up the SDK.

packages/analytics-js/src/constants/logMessages.ts
saikumarrs: I'd recommend a shorter message something like below, (not exact)

```
The input custom context value is not a plain object (or contains prototype pollution keys) that can be serialised to a JSON string.
```

packages/analytics-js/src/components/customContext/utilities.ts
dciccale: One side effect of this simplification is that a marker-only update to a missing path can leave an empty parent object. For example, `{ missing: { nested: null } }` applied to `{}` could result in `{ missing: {} }` instead of `{}`. If we are comfortable accepting that behavior, I’ll simplify the implementation and align the tests, HLD, and LLD accordingly.

packages/analytics-js/src/components/customContext/utilities.ts
dciccale: Agreed. We should keep the checks that protect the SDK, such as requiring a plain object and blocking prototype pollution. For other values, we can let normal event serialization handle them. I removed the extra value validation and updated the tests, HLD, and LLD.

packages/analytics-js/src/components/customContext/utilities.ts
saikumarrs: @dciccale I believe this is fine. We can document the behavior for the users.

dciccale: Updated. `prepareCustomContextUpdate` now keeps `null` and `undefined` markers. `CustomContextStore` merges the update and then calls `removeUndefinedAndNullValues`. The tests and specifications now document the accepted empty-parent behavior.

packages/analytics-js/src/constants/logMessages.ts
dciccale: Updated the warning to state the structural requirements directly. It now says: `Invalid custom context. Use a plain object without prototype pollution keys.` I left out the JSON serialization condition because nested values now follow the SDK’s normal serialization behavior.

saikumarrs (approved):

PR #3132, opened 2026-07-29
feat: sdk-5055 seed load-time custom context

dciccale:
## Summary

- extract `LoadOptions.context` without mutating caller-owned load options
- preserve raw context values until custom-context validation runs
- seed accepted context before the SDK lifecycle begins
- continue loading with an empty store when initial context is invalid
- rebase load seeding onto the shared application-state ownership model approved in the architecture review

## Validation

- 303 focused load, lifecycle, state, privacy, and integration tests passed on the rebased stack
- analytics-js and analytics-js-common lint passed with zero errors
- Prettier and `git diff --check` passed

Linear: [SDK-5055](https://linear.app/rudderstack/issue/SDK-5055/add-load-time-custom-context-support-to-js-sdk-load-options)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of custom context data provided during analytics loading.
  * Invalid context values are safely ignored, while valid data is sanitized and preserved.
  * Custom context is now correctly available throughout initialization and buffered event replay.
  * Load options are forwarded consistently without altering the original options object.

* **Tests**
  * Added coverage for load options, custom context handling, sanitization, lifecycle ordering, and invalid configurations.

### DIFF
--- a/packages/analytics-js/src/components/core/Analytics.ts
+++ b/packages/analytics-js/src/components/core/Analytics.ts
@@ -118,7 +118,12 @@ class Analytics implements IAnalytics {
   /**
    * Start application lifecycle if not already started
    */
-  load(writeKey: string, dataPlaneUrl: string, loadOptions: Partial<LoadOptions> = {}) {
+  load(
+    writeKey: string,
+    dataPlaneUrl: string,
+    loadOptions: Partial<LoadOptions> = {},
+    customContext?: unknown,
+  ) {
     if (state.lifecycle.status.value) {
       return;
     }
@@ -131,6 +136,10 @@ class Analytics implements IAnalytics {
     if (!isDataPlaneUrlValid(dataPlaneUrl)) {
       this.logger.error(DATA_PLANE_URL_VALIDATION_ERROR(ANALYTICS_CORE, dataPlaneUrl));
       return;
+    }
+
+    if (customContext !== undefined) {
+      this.customContextStore.set(customContext);
     }
 
     // Set initial state values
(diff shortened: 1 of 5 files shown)


packages/analytics-js/src/components/core/IAnalytics.ts
```diff
@@ -52,6 +52,7 @@ export interface IAnalytics {
     writeKey: string,
     dataPlaneUrl?: string | Partial<LoadOptions>,
     loadOptions?: Partial<LoadOptions>,
+    customContext?: unknown,
```
saikumarrs: Let's just simply take the custom context from the load options itself? This seems like extra work.
Are there any concerns?

dciccale: Agreed. I’ll keep `context` in `LoadOptions` through the `Analytics.load()` boundary, extract it there for `CustomContextStore`, and normalize only the remaining options. This removes the fourth argument while preserving the dedicated context validation and keeping customer context out of load-options state.

packages/analytics-js/src/components/core/Analytics.ts
```diff
@@ -133,11 +134,20 @@ class Analytics implements IAnalytics {
       return;
     }
 
+    const loadOptionsCopy = { ...loadOptions };
+    const customContext = loadOptionsCopy.context;
+    delete loadOptionsCopy.context;
+    const sanitizedLoadOptions = getSanitizedValue(loadOptionsCopy);
```
saikumarrs: Nitpick: Can we simplify to something like this? I'd avoid `delete` operation unless really required.

```suggestion
    const sanitizedLoadOptions = getSanitizedValue(clone(loadOptions));
    const customContext = loadOptions.context;
```

saikumarrs (dismissed):

packages/analytics-js/src/components/core/Analytics.ts
dciccale: Good point. I will replace `delete` with object rest destructuring. This keeps `context` on its custom-context validation path and out of `state.loadOptions`.

saikumarrs (dismissed):

saikumarrs (approved):

Issue #2967, opened 2026-04-22
BUG : useServerSideCookies: rsaRequest called on every page load with re-encrypted values, separate request per cookie


kgwood39:
1. With useServerSideCookies: true, /rsaRequest is POST'd on every page load even when all cookies already exist and are valid — expected behavior should be no request when cookies are current.

2. Each cookie generates a separate POST request rather than batching all cookies into a single request using the existing cookies array — results in 4 separate requests per page load

3. Cookie values are re-encrypted on every call producing a different ciphertext for the same underlying UUID — makes server-side change detection impossible and causes unnecessary Set-Cookie churn on every response.

Set these loadOptions for the rudderanalytics.load call:

var loadOptions = {
         // poor performing code
          useServerSideCookies: true,
          dataServiceEndpoint: "/rsaRequest"
      };

We expect to see on request for the all needed cookies, to have a useful cookie for it's lifetime and not re-encrypt and re-request these cookies with every page load.

This functionality is made to get around Safari ITP.  4 additional requests on every page load is too high a price to make this useful.

saikumarrs:
Hi @kgwood39, thanks for the detailed report and the clear reproduction steps — really appreciate you taking the time to surface this.

We're looking into the behavior you've described around `useServerSideCookies` and will follow up here with our findings and next steps shortly.

saikumarrs:
Hi @kgwood39 — thanks for the detailed writeup. I went through each of the three points; here's where we've landed.

### 1. Request fires even when cookies "look" valid

This is actually working as intended. Skipping the request when the value hasn't changed would keep the value fresh but let the TTL decay — the redundant-looking writes are deliberately renewing the cookie's server-issued lifetime on each load.

Also worth noting: `rl_session`'s payload legitimately differs every load because `sessionInfo.expiresAt` is bumped on each page view to keep the active session alive.

### 2. Separate request per cookie

This one we agree is a real inefficiency. `setServerSideCookies` already accepts a batch in a single payload — the caller is just invoking it once per key. Batching the page-load burst into one request is the target fix, and it's what actually moves the needle on the 4-requests-per-load number.

Treat this as in progress — there are a couple of in-flight-request edge cases we want to handle cleanly before landing. I'll follow up here once the PR is open.

### 3. Different ciphertext for the same underlying UUID

[...]

PR #3176, opened 2026-09-02
fix(analytics-js): skip no-op server-side cookie writes

saikumarrs:
## PR Description

With `useServerSideCookies: true`, every cookie-backed session key holding a value fires a `POST /rsaRequest` on **each page load** — typically four (`rl_anonymous_id`, `rl_session`, `rl_page_init_referrer`, `rl_page_init_referring_domain`). Three of those four send a value byte-identical to the cookie already stored.

`registerEffects()` registers one `effect()` per session key and they all run on init, so nothing distinguishes "this value changed" from "this value is exactly what is already persisted". This is the core of [#2967](https://github.com/rudderlabs/rudder-sdk-js/issues/2967), where the reporter turned the feature off because four extra requests per page load was too high a price for the Safari ITP workaround.

### The change

A guard at the top of the server-side-cookie branch in `syncValueToStorage`:

```ts
// Skip the request if the cookie already holds this exact value.
// The cookie is set with a long max age, so it needs no periodic renewal.
if (this.isValueAlreadyPersisted(curStore, cookieName, cookieValue)) {
  return;
}
```

`isValueAlreadyPersisted` reuses the same `stringifyWithoutCircular` comparison the response verification already performs in `setServerSideCookies`. Values are compared **after decryption**, so the check is unaffected by which encryption version a cookie was written with.

The comparison fails safe: string equality can only hold when the content is genuinely identical, so a differing key order causes an *unnecessary write* rather than a missed one. There is no input for which this skips a write it should have made.

### Why the redundant writes are not worth keeping

They were previously justified as renewing the cookie's server-issued TTL. `DEFAULT_COOKIE_MAX_AGE_MS` is **1 year**, and a genuine first-party cookie delivered via `Set-Cookie` is not subject to Safari ITP's 7-day cap — that cap applies to script-written cookies and CNAME-cloaked responses. Re-issuing a 1-year cookie on every page load buys nothing.

### Effect

Repeat page load: **4 requests → 1**. `rl_session` still writes every time, because `expiresAt` genuinely moves — which is deliberate and load-bearing. `refreshSession()` reads `sessionInfo` back from storage on every event and feeds it to `hasSessionExpired`; if the persisted `expiresAt` ever went stale, the next event would mint a **new session id** while the user was still active. A test pins that behaviour.

A first-time visitor still writes all four keys since they are genuinely new — that case is covered by the batching work in SDK-4981, so the two changes remain complementary.

**Out of scope:** throttling the per-event `sessionInfo` sync, and any new persisted state. Both are recorded on the Linear task with the reasoning.

## Linear task (optional)

[SDK-5409](https://linear.app/rudderstack/issue/SDK-5409/suppress-redundant-server-side-cookie-requests-in-the-js-sdk)

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

Not yet run for this change — the logic is storage-comparison only with no browser-specific API surface, but this is left unchecked deliberately rather than assumed.

## Sanity Suite

Not yet run.

## Security

## Test plan

Five new tests in a `redundant server-side cookie requests` block, written before the implementation:

- skips the request when the persisted value is already up to date
- still requests when the persisted value differs
- still requests when no cookie exists yet
- **still requests for `sessionInfo` as `expiresAt` changes** — the session-continuity regression guard
- does not skip the client-side write when server-side cookies are disabled

Verified locally:

- `userSessionManager` suite — 210 passed, 1 skipped
- `nx affected -t test` (the pre-commit gate) — passed across all 3 affected projects
- `check:size:build` and `check:size:json` — all bundles within limits
- `eslint` on changed files — 0 errors (2 warnings, both pre-existing)
- `tsc --noEmit` — no errors in the changed files

## Note for reviewers

The new tests restore `clientDataStoreCookie.set`/`.remove` in their `beforeEach`. Earlier tests in this file assign `jest.fn()` to those on the **shared store singleton** and never restore them, so any later test that tries to seed a real cookie silently writes nothing. That cost me a false green before I caught it. Worked around locally rather than refactoring the file — worth a separate cleanup.

Also deliberate: skipping does **not** cancel an already-scheduled debounce timer. If a value flips back to the persisted one inside the 10 ms window, that pending request still fires — harmless and rare. Cancelling it correctly also requires clearing `serverSideCookiesRequestInProgress`, which belongs with the separate `reset()`-race bug (removal path at `UserSessionManager.ts:588` clears neither), tracked separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented server-side cookie synchronization from being skipped when a matching request is already pending or when the cookie was rewritten during storage migration.
  * Preserved synchronization after an earlier response clears the in-progress state, improving consistency across repeated updates.

* **Tests**
  * Added coverage for pending requests, storage migration, repeated synchronization, and cleanup of shared test state.

### DIFF
--- a/packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
+++ b/packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
@@ -2683,6 +2683,59 @@ describe('User session manager', () => {
         expect(setServerSideCookiesSpy).toHaveBeenCalledTimes(1);
       });
 
+      // Storage migration rewrites the cookie client side. If the guard then skips, the cookie
+      // stays a script written cookie for good, and Safari ITP caps it at 7 days - which is the
+      // exact outcome server-side cookies exist to avoid.
+      it('should make a request for a cookie that storage migration rewrote', () => {
+        // Only stub the migration extension point. The store uses the same plugins manager
+        // for encryption, so a blanket mock would write the cookie unencrypted.
+        const originalInvokeSingle = defaultPluginsManager.invokeSingle.bind(defaultPluginsManager);
+        const migrateSpy = jest
+          .spyOn(defaultPluginsManager, 'invokeSingle')
+          .mockImplementation((extPoint?: string, ...args: any[]) =>
+            extPoint === 'storage.migrate'
+              ? dummyAnonymousId
+              : originalInvokeSingle(extPoint, ...args),
+          );
+        state.storage.migrate.value = true;
+        state.session.anonymousId.value = dummyAnonymousId;
+        const setServerSideCookiesSpy = jest.spyOn(userSessionManager, 'setServerSideCookies');
+
+        // Migration writes the value into the cookie client side
+        userSessionManager.migrateStorageIfNeeded(undefined, ['anonymousId']);
+        expect(clientDataStoreCookie.get(COOKIE_KEYS.anonymousId)).toEqual(dummyAnonymousId);
+
+        userSessionManager.syncValueToStorage('anonymousId');
+        jest.advanceTimersByTime(1000);
+
+        expect(setServerSideCookiesSpy).toHaveBeenCalled();
+        migrateSpy.mockRestore();
+      });
+
+      // With two writes in flight for one key, the earlier response clears the shared
+      // in-progress flag while the later one is still outstanding. Skipping on the strength
+      // of that flag would let the later write land last and diverge the cookie from state.
+      it('should keep requesting after an earlier response clears the in-progress flag', () => {
+        clientDataStoreCookie.set(COOKIE_KEYS.anonymousId, dummyAnonymousId);
+        const setServerSideCookiesSpy = jest.spyOn(userSessionManager, 'setServerSideCookies');
+
+        state.session.anonymousId.value = 'pending_anonymousId';
+        userSessionManager.syncValueToStorage('anonymousId');
+        jest.advanceTimersByTime(1000);
+        setServerSideCookiesSpy.mockClear();
+
+        // An earlier response lands and clears the shared flag, even though a write is
+        // still outstanding for this key
+        userSessionManager.serverSideCookiesRequestInProgress.anonymousId = false;
+
+        // State returns to the value the cookie still holds
+        state.session.anonymousId.value = dummyAnonymousId;
+        userSessionManager.syncValueToStorage('anonymousId');
+        jest.advanceTimersByTime(1000);
+
+        expect(setServerSideCookiesSpy).toHaveBeenCalledTimes(1);
+      });
+
       it('should make a request if the cookie does not exist yet', () => {
         state.session.anonymousId.value = dummyAnonymousId;
         const setServerSideCookiesSpy = jest.spyOn(userSessionManager, 'setServerSideCookies');
--- a/packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
+++ b/packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
@@ -2630,6 +2630,10 @@ describe('User session manager', () => {
       });
 
       afterEach(() => {
+        // The store is a shared singleton, so any override made here would otherwise leak
+        // into later tests. Drop them before handing the timers back.
+        delete (clientDataStoreCookie as Partial<Store>).set;
+        delete (clientDataStoreCookie as Partial<Store>).remove;
         jest.useRealTimers();
       });
 
@@ -2646,6 +2650,20 @@ describe('User session manager', () => {
 
       it('should make a request if the persisted value is different', () => {
         clientDataStoreCookie.set(COOKIE_KEYS.anonymousId, 'stale_anonymousId');
+        state.session.anonymousId.value = dummyAnonymousId;
+        const setServerSideCookiesSpy = jest.spyOn(userSessionManager, 'setServerSideCookies');
+
+        userSessionManager.syncValueToStorage('anonymousId');
+        jest.advanceTimersByTime(1000);
+
+        expect(setServerSideCookiesSpy).toHaveBeenCalled();
+      });
+
+      // An A -> B -> A transition: the B request is still pending when the value reverts to A.
+      // Skipping here would let B land afterwards, leaving the cookie and the state divergent.
+      it('should make a request if one is already pending for the same key', () => {
+        clientDataStoreCookie.set(COOKIE_KEYS.anonymousId, dummyAnonymousId);
+        userSessionManager.serverSideCookiesRequestInProgress.anonymousId = true;
         state.session.anonymousId.value = dummyAnonymousId;
         const setServerSideCookiesSpy = jest.spyOn(userSessionManager, 'setServerSideCookies');
 
--- a/packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
+++ b/packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
@@ -2616,6 +2616,91 @@ describe('User session manager', () => {
         done();
       }, 1000);
     });
+
+    describe('redundant server-side cookie requests', () => {
+      beforeEach(() => {
+        jest.useFakeTimers();
+        // Earlier tests replace these on the shared store instance and never restore them.
+        // Drop the own properties so the real prototype implementations are used again.
+        delete (clientDataStoreCookie as Partial<Store>).set;
+        delete (clientDataStoreCookie as Partial<Store>).remove;
+        state.serverCookies.isEnabledServerSideCookies.value = true;
+        state.storage.entries.value = entriesWithOnlyCookieStorage;
+        state.serverCookies.dataServiceUrl.value = 'https://dummy.dataplane.host.com/rsaRequest';
+      });
+
+      afterEach(() => {
+        jest.useRealTimers();
+      });
+
+      it('should not make a request if the persisted value is already up to date', () => {
+        clientDataStoreCookie.set(COOKIE_KEYS.anonymousId, dummyAnonymousId);
+        state.session.anonymousId.value = dummyAnonymousId;
+        const setServerSideCookiesSpy = jest.spyOn(userSessionManager, 'setServerSideCookies');
+
+        userSessionManager.syncValueToStorage('anonymousId');
+        jest.advanceTimersByTime(1000);
+
+        expect(setServerSideCookiesSpy).not.toHaveBeenCalled();
+      });
+
+      it('should make a request if the persisted value is different', () => {
+        clientDataStoreCookie.set(COOKIE_KEYS.anonymousId, 'stale_anonymousId');
+        state.session.anonymousId.value = dummyAnonymousId;
+        const setServerSideCookiesSpy = jest.spyOn(userSessionManager, 'setServerSideCookies');
+
+        userSessionManager.syncValueToStorage('anonymousId');
+        jest.advanceTimersByTime(1000);
+
+        expect(setServerSideCookiesSpy).toHaveBeenCalled();
+      });
+
+      it('should make a request if the cookie does not exist yet', () => {
+        state.session.anonymousId.value = dummyAnonymousId;
+        const setServerSideCookiesSpy = jest.spyOn(userSessionManager, 'setServerSideCookies');
+
+        userSessionManager.syncValueToStorage('anonymousId');
+        jest.advanceTimersByTime(1000);
+
+        expect(setServerSideCookiesSpy).toHaveBeenCalled();
+      });
+
+      // The session cookie must keep syncing, otherwise the persisted expiresAt goes stale and
+      // the next event would start a new session while the user is still active.
+      it('should still make a request for the session info as expiresAt changes', () => {
+        const persistedSessionInfo = {
+          id: 1234567890,
+          expiresAt: 1234567890 + DEFAULT_SESSION_TIMEOUT_MS,
+          autoTrack: true,
+          timeout: DEFAULT_SESSION_TIMEOUT_MS,
+        };
+        clientDataStoreCookie.set(COOKIE_KEYS.sessionInfo, persistedSessionInfo);
+        state.session.sessionInfo.value = {
+          ...persistedSessionInfo,
+          expiresAt: persistedSessionInfo.expiresAt + 60 * 1000,
+        };
+        const setServerSideCookiesSpy = jest.spyOn(userSessionManager, 'setServerSideCookies');
+
+        userSessionManager.syncValueToStorage('sessionInfo');
+        jest.advanceTimersByTime(1000);
+
+        expect(setServerSideCookiesSpy).toHaveBeenCalled();
+      });
+
+      it('should not skip the client-side write when server-side cookies are disabled', () => {
+        state.serverCookies.isEnabledServerSideCookies.value = false;
+        clientDataStoreCookie.set(COOKIE_KEYS.anonymousId, dummyAnonymousId);
+        clientDataStoreCookie.set = jest.fn();
+        state.session.anonymousId.value = dummyAnonymousId;
+
+        userSessionManager.syncValueToStorage('anonymousId');
+
+        expect(clientDataStoreCookie.set).toHaveBeenCalledWith(
+          COOKIE_KEYS.anonymousId,
+          dummyAnonymousId,
+        );
+      });
+    });
   });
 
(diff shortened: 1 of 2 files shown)


packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
```diff
@@ -2616,6 +2616,91 @@ describe('User session manager', () => {
...
+        delete (clientDataStoreCookie as Partial<Store>).set;
+        delete (clientDataStoreCookie as Partial<Store>).remove;
+        state.serverCookies.isEnabledServerSideCookies.value = true;
+        state.storage.entries.value = entriesWithOnlyCookieStorage;
+        state.serverCookies.dataServiceUrl.value = 'https://dummy.dataplane.host.com/rsaRequest';
+      });
+
+      afterEach(() => {
+        jest.useRealTimers();
+      });
```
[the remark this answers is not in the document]
saikumarrs: Good catch — fixed in 8f6eec64.

The last test in the block did reassign `clientDataStoreCookie.set` without restoring it, and since the store is a shared singleton that would leak into later describes. The block now drops both overridden own properties in `afterEach`, before handing the timers back, so the prototype implementations are restored regardless of which test ran.

Worth noting this is pre-existing in the file rather than new: several earlier tests assign `jest.fn()` to the same singleton and never restore it, which is why the `beforeEach` here has to delete them in the first place — a later test that seeds a real cookie otherwise writes nothing silently. That cost me a false green before I caught it. I've kept the workaround local to this block rather than refactoring the whole file, but it's worth a separate cleanup.

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
```diff
@@ -553,6 +571,12 @@ class UserSessionManager implements IUserSessionManager {
           state.serverCookies.isEnabledServerSideCookies.value &&
           storageType === COOKIE_STORAGE
         ) {
+          // Skip the request if the cookie already holds this exact value.
+          // The cookie is set with a long max age, so it needs no periodic renewal.
+          if (this.isValueAlreadyPersisted(curStore, cookieName, cookieValue)) {
+            return;
```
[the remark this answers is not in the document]
saikumarrs: Valid, and the underlying problem is broader than the timer — fixed in 8f6eec64, though not quite as suggested.

Cancelling the pending timer alone isn't sufficient, because the same divergence happens with no timer involved. If a request for B is already **in flight** and the state reverts to A (the value the cookie still holds), the guard reads the cookie, sees A, and skips — then B's response lands and sets the cookie to B while the state says A, with nothing to reconcile them. Clearing `serverSideCookiesRequestInProgress` on skip would make that case worse, since the in-flight B write would then be entirely unaccounted for.

So rather than tracking timer completion separately from in-flight state, the guard simply doesn't skip while anything is pending for that key:

```ts
if (
  !this.serverSideCookiesRequestInProgress[sessionKey] &&
  this.isValueAlreadyPersisted(curStore, cookieName, cookieValue)
) {
  return;
}
```

An A → B → A transition now falls through, clears the pending timer and schedules the write — exactly the behaviour before this guard existed. The optimisation still applies on page load, where nothing is pending, which is the case the change is actually for.

Regression test added as suggested: `should make a request if one is already pending for the same key`. I verified it genuinely fails with the `serverSideCookiesRequestInProgress` condition removed, rather than passing vacuously.

packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
```diff
@@ -2616,6 +2616,109 @@ describe('User session manager', () => {
...
+
+        expect(setServerSideCookiesSpy).toHaveBeenCalled();
+      });
+
+      // An A -> B -> A transition: the B request is still pending when the value reverts to A.
+      // Skipping here would let B land afterwards, leaving the cookie and the state divergent.
+      it('should make a request if one is already pending for the same key', () => {
+        clientDataStoreCookie.set(COOKIE_KEYS.anonymousId, dummyAnonymousId);
+        userSessionManager.serverSideCookiesRequestInProgress.anonymousId = true;
+        state.session.anonymousId.value = dummyAnonymousId;
```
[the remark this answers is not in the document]
saikumarrs: Fair — the test asserted the guard's precondition rather than the transition its comment described. Rewritten in 7d8e6f62.

It now drives the sequence through the real path: B is written via `syncValueToStorage`, the debounce is allowed to elapse so the request is genuinely in flight and unresolved, and only then does the value revert to A. `serverSideCookiesRequestInProgress` is no longer set by hand — the production code sets it, and the test asserts it.

```ts
state.session.anonymousId.value = 'pending_anonymousId';
userSessionManager.syncValueToStorage('anonymousId');
jest.advanceTimersByTime(1000);

expect(setServerSideCookiesSpy).toHaveBeenCalledTimes(1);
expect(userSessionManager.serverSideCookiesRequestInProgress.anonymousId).toBe(true);
setServerSideCookiesSpy.mockClear();

// Back to A while B is still unresolved
state.session.anonymousId.value = dummyAnonymousId;
userSessionManager.syncValueToStorage('anonymousId');
jest.advanceTimersByTime(1000);

expect(setServerSideCookiesSpy).toHaveBeenCalledTimes(1);
```

Letting the debounce elapse before the revert matters: if B were still only queued, both the guarded and unguarded code would end up issuing exactly one request, and the test would pass vacuously. Waiting until B is in flight is what makes the skip observable. I re-checked that it fails with the `serverSideCookiesRequestInProgress` condition removed.

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
```diff
@@ -553,6 +571,19 @@ class UserSessionManager implements IUserSessionManager {
...
         ) {
+          // Skip the request if the cookie already holds this exact value.
+          // The cookie is set with a long max age, so it needs no periodic renewal.
+          //
+          // Only safe while nothing is pending for this key. A queued or in-flight request
+          // may still write a different value, so skipping then would leave the cookie and
+          // the state divergent with nothing to reconcile them.
+          if (
+            !this.serverSideCookiesRequestInProgress[sessionKey] &&
+            this.isValueAlreadyPersisted(curStore, cookieName, cookieValue)
```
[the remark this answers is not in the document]
saikumarrs: Tracked separately as [SDK-5411](https://linear.app/rudderstack/issue/SDK-5411/server-side-cookie-writes-are-not-serialised-per-key) rather than fixed here, because it is pre-existing on `develop` and unrelated to this change. Worth recording where I agree and where the described mechanism does not match the code.

**The stated mechanism does not hold.** A late-arriving callback does not apply its own snapshot. `getCurrentCookieValuesFromState()` is evaluated *inside* the callback, so `cb(cData.name, cData.value)` writes the **current** state value, not the value that request was built from. No stale value is restored through that path.

**But the conclusion is right, one level down.** The ordering hazard is on the wire. The browser applies `Set-Cookie` from whichever response lands last, so if the older request's response arrives second, the cookie ends up on the older value — and the verification block cannot catch it, because each callback compares its *own* expectation against the cookie:

```
Req A (value B) sent, Req B (value C) sent
Resp C arrives → cookie = C. Callback: expected C, current C → match, no correction
Resp B arrives → cookie = B. Callback: expected B, current B → match, no correction
Final: cookie = B, state = C
```

Both callbacks see a match, so the divergence is silent.

**The flag point is correct.** `clearInProgressFlags()` is a single boolean per key, so whichever request finishes first clears state a newer in-flight request still owns.

Scope call: this PR adds a one-line skip guard, and the defect above reproduces on `develop` without it. The guard does consult that flag, so an inaccurate flag narrows the window it protects — noted on the ticket as a reason to fix the two together. Serialising writes per key or tracking request generations is the fix, along with deferred-callback tests for both response orderings, which is more than this change should carry.

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
```diff
@@ -553,6 +571,19 @@ class UserSessionManager implements IUserSessionManager {
...
           storageType === COOKIE_STORAGE
         ) {
+          // Skip the request if the cookie already holds this exact value.
+          // The cookie is set with a long max age, so it needs no periodic renewal.
+          //
+          // Only safe while nothing is pending for this key. A queued or in-flight request
+          // may still write a different value, so skipping then would leave the cookie and
+          // the state divergent with nothing to reconcile them.
+          if (
+            !this.serverSideCookiesRequestInProgress[sessionKey] &&
```
dciccale: I think there is still an overlap case here. If B and C requests are pending, B can finish first and clear the flag while C remains active. If state then returns to stored B, this guard returns, while the base code would schedule another write. C can then finish and leave the cookie at C while state is B. The current test keeps only one request pending, so it does not cover this response order. Could you check this case?

saikumarrs: You're right, and this is a sharper point than the one I answered earlier in the thread above. Confirming the trace: `clearInProgressFlags()` clears the flag for the keys in *that* request's map, so B's callback clears it while C is still in flight. The guard then reads a flag that is wrongly `false`, sees the persisted value match, and returns.

The part I had wrong: I'd written on the follow-up ticket that this "does not create a new class of bug." It does. Without the guard, that step schedules a write for B and the cookie has a chance to converge back. With the guard, that write is never scheduled — so the interleaving goes from probabilistically convergent to **deterministically** divergent. And you're right that the tests here never reach it, since they only ever hold one request pending.

We've decided not to fix it in this PR. The root cause is that `serverSideCookiesRequestInProgress` is a single shared boolean per key, and making it accurate is the concurrency rework tracked in [SDK-5411](https://linear.app/rudderstack/issue/SDK-5411/server-side-cookie-writes-are-not-serialised-per-key) — serialising writes per key or tracking request generations. Doing a partial version of that here would mean carrying half a concurrency model in a PR whose actual change is a one-line guard.

I've updated SDK-5411 with your trace verbatim, corrected the claim above, and added a required test for exactly this ordering: two writes in flight for one key, the **older** response arriving first, state then reverting to what the cookie holds, asserting the final cookie matches the final state.

Considered and rejected: gating the guard on a monotonic "has a write ever been issued for this key in this page load" flag instead of the in-progress one. It would close this specific hole in about three lines, since a flag that is never cleared can't be wrongly false. We'd rather SDK-5411 fix the underlying flag properly than have this PR work around it — but say the word if you'd prefer the interim guard.

Leaving this thread open for you rather than resolving it, since it's acknowledged rather than fixed.

packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
```diff
@@ -2616,6 +2616,119 @@ describe('User session manager', () => {
...
+
+      afterEach(() => {
+        // The store is a shared singleton, so any override made here would otherwise leak
+        // into later tests. Drop them before handing the timers back.
+        delete (clientDataStoreCookie as Partial<Store>).set;
+        delete (clientDataStoreCookie as Partial<Store>).remove;
+        jest.useRealTimers();
+      });
+
+      it('should not make a request if the persisted value is already up to date', () => {
```
dciccale, line 2640: The migration path appears safe because `syncStorageDataToState()` runs `migrateStorageIfNeeded()` before `registerEffects()`. However, SDK-5409 asks for explicit coverage of this interaction. Could we add a test that seeds a legacy-encrypted cookie, runs initialization, and verifies migration occurs before the equal-value guard skips the server request?

saikumarrs: Added in aa25bfc3 — and writing it turned up a real bug, so thank you for pushing on it.

Your ordering analysis is right: `migrateStorageIfNeeded()` runs inside `syncStorageDataToState()`, before `registerEffects()`, so no legacy ciphertext survives to the guard. That part is safe.

The problem is *how* migration rewrites the cookie. `migrateStorageIfNeeded` calls `store.set(...)` — a **client-side** write. So:

1. Legacy cookie migrated, rewritten by script
2. State set to the same value
3. Effect fires, guard sees persisted == state, **skips the server request**

Before this PR, step 3 issued a server-side write that promoted the cookie from script-set to server-set. With the guard it doesn't, and since the value never changes it is skipped on every subsequent load too. The cookie stays script-written for good — which Safari ITP caps at 7 days, after which `rl_anonymous_id` expires and a new anonymous ID is generated. Precisely the outcome this feature exists to prevent.

It generalises beyond migration: JavaScript cannot tell whether a cookie was set by the server or by script, so the guard also could not re-promote a cookie left behind by the client-side fallback after a failed request.

Fix is to track the keys the SDK has written during this page load — set by `migrateStorageIfNeeded`, by `migrateDataFromPreviousStorage`, and when a server-side request is issued — and never skip those. It is never cleared, so it cannot go stale.

That also replaces the in-progress check in the guard, which closes [your other comment](https://github.com/rudderlabs/rudder-sdk-js/pull/3176#discussion_r3912978561). The monotonic flag is true for any key with an outstanding write, so it subsumes that check while being immune to `clearInProgressFlags()` clearing state a newer request still owns.

Two tests added, both confirmed to fail against the previous guard:
* `should make a request for a cookie that storage migration rewrote`
* `should keep requesting after an earlier response clears the in-progress flag`

The deeper concurrency issue behind your other comment is still real and still tracked in [SDK-5411](https://linear.app/rudderstack/issue/SDK-5411/server-side-cookie-writes-are-not-serialised-per-key) — the flag stops the guard from making it worse, but it does not serialise the writes.

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
dciccale: The monotonic flag addresses the overlap introduced by this guard. Keeping the wider concurrency work in SDK-5411 makes sense. I am resolving this thread.

packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
dciccale: The new test covers the client-side migration write. The monotonic flag also prevents the equal-value guard from skipping the server request. This addresses my migration concern. I am resolving this thread.

packages/analytics-js/src/components/userSessionManager/UserSessionManager.ts
```diff
@@ -95,6 +95,13 @@ class UserSessionManager implements IUserSessionManager {
    * Tracks whether a server-side cookie setting request is in progress or not.
    */
   serverSideCookiesRequestInProgress: Record<UserSessionKey, boolean>;
+  /**
+   * Tracks the keys whose cookie the SDK has written during this page load, either
+   * client-side or through the data service. Once set, the value in the cookie can no
+   * longer be assumed to have been set by the data service, so the equal-value skip in
+   * syncValueToStorage no longer applies. Never cleared.
+   */
+  serverSideCookieSyncRequired: Record<UserSessionKey, boolean>;
```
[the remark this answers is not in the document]
saikumarrs: Valid, and accepted as a documented limitation rather than fixed here. Recorded on [SDK-5409](https://linear.app/rudderstack/issue/SDK-5409/suppress-redundant-server-side-cookie-requests-in-the-js-sdk) with the reasoning below.

Confirming the trace (the field is now named `cookieWrittenInPageLoad` after 2ef912bd, but the analysis is unchanged): a fallback write in page load 1 leaves a script-written cookie, page load 2 starts with an empty record and an unchanged value, and the guard skips. The cookie is never promoted back to server-set.

Worth noting what *is* covered, since it is the same bug one level in: within a single page load, both storage migration and the client-side fallback set the flag, so neither can be skipped. The fallback case specifically is covered because the flag is set when the request is issued, and a fallback only ever follows a request. What a per-page-load flag cannot see is evidence from a *previous* page load.

You are right that fixing it needs a durable marker, and as you note that conflicts with the ticket's constraint of adding no new persisted data — which exists because it drags in storage-availability, quota and consent handling for what is otherwise a stateless comparison.

Accepting it on these grounds:

* It needs a failed or mismatched request to trigger at all.
* On Safari the script-written cookie is ITP-capped at 7 days, so it expires, the value is regenerated, and the next write goes through the data service and promotes the cookie. It self-heals, at the cost of one anonymous ID reset for affected users.
* On browsers without the cap there is no practical harm — the cookie keeps its full lifetime either way.

If we revisit it, the ticket records the shape: a durable per-key promotion marker set on fallback writes, cleared only after a successful server-side write, plus the reload regression test you describe. That belongs in its own piece of work alongside the storage-permission handling, not bolted onto this change.

dciccale (approved):

PR #3177, opened 2026-09-02
fix(analytics-js): serialise server-side cookie writes per key

saikumarrs:
> **Stacked on #3176.** Base is `feature/sdk-5409-…`, not `develop`. Review #3176 first; this diff is only the serialisation change on top of it.

## PR Description

Each `POST /rsaRequest` is issued with no ordering guarantee. The 10 ms debounce in `syncValueToStorage` only collapses changes inside its own window, so two changes further apart put **two requests on the wire for the same cookie**. The browser applies `Set-Cookie` in response arrival order, so if the older response lands second the cookie settles on the older value:

```
Req A (value B) sent, Req B (value C) sent
Resp C arrives → cookie = C. Callback: expected C, current C → match, no correction
Resp B arrives → cookie = B. Callback: expected B, current B → match, no correction
Final: cookie = B, state = C
```

The verification block cannot catch this. Each callback compares its **own request-time snapshot** against the cookie, so both see a match and neither corrects. The divergence is silent and permanent until the next write to that key.

Pre-existing on `develop` — surfaced by review on #3176, where the skip guard made one interleaving of it deterministic rather than merely possible. That specific interaction was fixed in #3176 itself; this PR fixes the underlying defect.

### Why serialise rather than track generations

Generation tracking can only *detect* a stale response — by the time a callback runs the browser has already applied that `Set-Cookie`, and it cannot be un-applied. Only repaired, and repair means a client-side write that Safari ITP caps at 7 days. Serialising prevents the reordering outright, which is what the acceptance criterion asks for.

### The change

The single in-progress boolean conflated two states. Splitting them is what makes the rest fall out:

- `serverSideCookiesInFlight` — request actually on the wire
- `serverSideCookiesResyncPending` — value changed while in flight

While a request is in flight for a key, a further change is **held back rather than sent concurrently**, and reissued against the then-current value once the response lands. Only one request per key is ever on the wire.

Two things follow for free:

1. **The in-progress flag becomes accurate.** With a single owner per key, `clearInProgressFlags` can no longer clear state a newer request still owns — which is what made `getUserSessionValue` fall back to storage while a write was outstanding.
2. **The removal path stops resurrecting cookies.** It now drops a queued write before removing, and if a request is already in flight it re-applies the removal once the response lands. A `reset()` racing a write can no longer bring the cookie back as an encrypted empty string.

`completeRequest()` also moved to the **end** of the response callback, after any client-side correction, so a resync reissued from it sees the settled cookie value.

## Linear task (optional)

[SDK-5411](https://linear.app/rudderstack/issue/SDK-5411/server-side-cookie-writes-are-not-serialised-per-key)

## Cross Browser Tests

- [ ] Chrome
- [ ] Firefox
- [ ] IE11

Not run. The change is request-sequencing logic with no browser-specific API surface, but left unchecked rather than assumed.

## Sanity Suite

Not run.

## Security

## Test plan

Five new tests in a `serialised server-side cookie writes` block, using a deferred-callback harness so response timing is controlled rather than raced:

- no second request while one is in flight for the same key
- the held-back write is issued once the in-flight response lands
- the in-progress flag stays set while a resync is pending
- a queued write does not resurrect a removed cookie
- a removal during an in-flight request is re-applied after the response

Verified: 218 passed / 1 skipped, `nx affected -t test` green across all 3 affected projects, eslint 0 errors on changed files.

## Note for reviewers

Two tests from #3176 changed assertions here. They asserted a second request was issued immediately while one was in flight — which is exactly the behaviour this PR fixes. They now assert the write is **held for reissue** rather than lost, which is the invariant that actually matters: `serverSideCookiesResyncPending` is set instead of the write being dropped.

## Follow-up

SDK-4981 (batching) will rework this same pending machinery again — a batch covers several keys, so per-key in-flight state collapses into one pending map plus a single in-flight request. Merge order matters: #3176, then this, then batching.

### DIFF
--- a/packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
+++ b/packages/analytics-js/__tests__/components/userSessionManager/UserSessionManager.test.ts
@@ -2683,6 +2683,59 @@ describe('User session manager', () => {
         expect(setServerSideCookiesSpy).toHaveBeenCalledTimes(1);
       });
 
+      // Storage migration rewrites the cookie client side. If the guard then skips, the cookie
+      // stays a script written cookie for good, and Safari ITP caps it at 7 days - which is the
+      // exact outcome server-side cookies exist to avoid.
+      it('should make a request for a cookie that storage migration rewrote', () => {
+        // Only stub the migration extension point. The store uses the same plugins manager
+        // for encryption, so a blanket mock would write the cookie unencrypted.
+        const originalInvokeSingle = defaultPluginsManager.invokeSingle.bind(defaultPluginsManager);
+        const migrateSpy = jest
+          .spyOn(defaultPluginsManager, 'invokeSingle')
+          .mockImplementation((extPoint?: string, ...args: any[]) =>
+            extPoint === 'storage.migrate'
+              ? dummyAnonymousId
+              : originalInvokeSingle(extPoint, ...args),
+          );
+        state.storage.migrate.value = true;
+        state.session.anonymousId.value = dummyAnonymousId;
+        const setServerSideCookiesSpy = jest.spyOn(userSessionManager, 'setServerSideCookies');
+
+        // Migration writes the value into the cookie client side
+        userSessionManager.migrateStorageIfNeeded(undefined, ['anonymousId']);
+        expect(clientDataStoreCookie.get(COOKIE_KEYS.anonymousId)).toEqual(dummyAnonymousId);
+
+        userSessionManager.syncValueToStorage('anonymousId');
+        jest.advanceTimersByTime(1000);
+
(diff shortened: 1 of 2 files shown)


saikumarrs:
Superseded by #3179 and closing.

This branch was cut before #3178 (batching), which replaces the per-key debounce timers with one shared timer and a pending map. The serialisation here was built on the per-key model, so most of its diff no longer applies — and two of its commits touch #3176 code that #3178 reverts. Rebasing would have left a diff that misleads more than it explains.

#3179 is the same design and the same tests, re-implemented on top of batching, where it comes out considerably smaller: the pending map is already the queue, so the per-key in-flight and resync records collapse into a single boolean.

PR #3210, opened 2026-09-15
feat(analytics-js-integrations): customer.io v2 client

ItsSudip:
## PR Description

Adds support for Customer.io's JavaScript client (`cioanalytics`, loaded from `cdp.customer.io` with a Data Pipelines write key) to the Customer.io device-mode integration, alongside the existing legacy snippet (`_cio`, Site ID). The client is selected per destination by the new `sdkVersion` config value:

- `'v2'` loads the JavaScript client. `init()` aborts with an error if `writeKey` is empty. The RudderStack anonymous ID is applied via `cioanalytics.setAnonymousId` before any event. The In-App Plugin load option is passed only when `anonymousInApp` is enabled; `siteId` is never passed (Customer.io supplies it from the write key's settings).
- Anything else, including a missing key, loads the legacy snippet exactly as before.

`identify`, `track`, `page`, `isLoaded` branch on the version (`identify(userId, traits)` on v2 vs `identify({ id, ...traits })` on v1). New `utils.js` with `getSdkVersion`; the loader is split into `loadNativeSdkV1` (body unchanged) and `loadNativeSdkV2`. New `browser.test.js` with 40 cases covering version resolution, both loaders for US/EU, the anonymous-ID and plugin-option behaviour, the empty-write-key path, and all event methods on both versions.

Bundle size: Customer.io bundle grows ~0.6 KB gzipped; all-integrations limits pass.

## Linear task (optional)

https://linear.app/rudderstack/issue/INT-7131

## Anonymous in-app message test:

https://github.com/user-attachments/assets/b30100ac-3c1f-4d00-8fb8-55e6edff9740

## Cross Browser Tests

- [x] Chrome
- [ ] Firefox
- [ ] IE11

## Sanity Suite

## Security

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0115NSHUVD6QT1KreXEeJMji

## Summary by CodeRabbit

- **New Features**
  - Added support for Customer.io SDK v2 alongside the existing v1 integration.
  - Added SDK version and US/EU data center configuration.
  - Added Data Pipelines loading with write key support.
  - Added optional anonymous in-app messaging and anonymous ID synchronization.
  - Added native Customer.io client access.
  - Improved identify, track, and page event forwarding across both SDK versions.

- **Bug Fixes**
  - Customer.io SDK v2 initialization now clearly fails when a write key is missing.


packages/analytics-js-integrations/src/integrations/CustomerIO/browser.js
```diff
@@ -28,45 +38,101 @@ class CustomerIO {
...
+    loadNativeSdkV1(siteID, datacenter, dataUseInApp);
+  }
+
+  initV2() {
+    const { writeKey, datacenter, anonymousInApp, analytics } = this;
+    if (!writeKey) {
+      logger.error(
+        'writeKey is required to load the Customer.io JavaScript client (SDK version 2.x); aborting load',
+      );
+      return;
```
saikumarrs: Let's throw an except from here.

packages/analytics-js-integrations/src/integrations/CustomerIO/browser.js
```diff
@@ -28,45 +38,101 @@ class CustomerIO {
...
-    const { traits } = context || {};
+    const traits = context?.traits ?? {};
     if (!userId) {
       logger.error('userId is required for Identify call');
       return;
     }
+    const client = this.getNativeClient();
+    if (!client) {
+      return;
+    }
```
saikumarrs: We don't need this check in any of the APIs given that isLoaded check already ensures it exists.
Moreover, it is a silent failure that'd be hard to debug.

packages/analytics-js-integrations/src/integrations/CustomerIO/browser.js
ItsSudip: Done in e8ef28f2 — `initV2()` now throws when `writeKey` is missing instead of logging and returning. The device-mode framework wraps `init()` in a try/catch, so the destination is marked failed and the error surfaced.

packages/analytics-js-integrations/src/integrations/CustomerIO/browser.js
ItsSudip: Done in e8ef28f2 — removed the per-API `!client` guards from `identify`/`track`/`page`. Event methods run only after `isReady()`, so the checks were dead code and a silent failure path.

packages/analytics-js-integrations/src/integrations/CustomerIO/browser.js
```diff
@@ -28,45 +38,101 @@ class CustomerIO {
...
   isReady() {
     return this.isLoaded();
   }
 
+  /**
+   * The loaded client for the configured version. Undefined only if `init()` aborted (2.x
+   * without a write key); `isReady()` is false then, so callers treat it as a no-op.
+   */
+  getNativeClient() {
+    return this.sdkVersion === SDK_V2 ? window[V2_GLOBAL_NAME] : window._cio;
```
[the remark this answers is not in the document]
ItsSudip: Addressed in e8ef28f2: `initV2()` now throws when `writeKey` is missing, so an aborted destination is marked failed and never becomes ready — it can't forward events off a pre-existing `cioanalytics` global. (Two web-device Customer.io destinations with different write keys is inherent to the shared `cioanalytics` singleton; the loader already no-ops a second `load`.)

packages/analytics-js-integrations/src/integrations/CustomerIO/browser.js
```diff
@@ -1,8 +1,7 @@
-/* eslint-disable no-underscore-dangle */
-/* eslint-disable class-methods-use-this */
-import { NAME, DISPLAY_NAME } from './constants';
+import { NAME, DISPLAY_NAME, SDK_V2, V2_GLOBAL_NAME, IN_APP_PLUGIN_KEY } from './constants';
```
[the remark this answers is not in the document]
ItsSudip: `no-underscore-dangle` isn't enabled in this package's ESLint config — `npx eslint` on this file passes (exit 0), and `--print-config` shows the rule unset. So `window._cio` doesn't fail lint and the removed directive was flagged as unused. No change needed.

packages/analytics-js-integrations/src/integrations/CustomerIO/nativeSdkLoader.js
```diff
@@ -30,4 +43,49 @@ function loadNativeSdk(siteID, datacenter, dataUseInApp) {
...
+  analytics.factory = method =>
+    function () {
+      analytics.push([method].concat(Array.prototype.slice.call(arguments, 0)));
+      return analytics;
+    };
+  analytics.methods.forEach(method => {
+    analytics[method] = analytics.factory(method);
+  });
+  analytics.load = (key, options) => {
+    const scriptElement = document.createElement('script');
```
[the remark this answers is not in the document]
ItsSudip: Device-mode integrations are browser-only: the service-worker package doesn't import `@rudderstack/analytics-js-integrations` (as noted), and every device-mode integration accesses `window`/`document`. Guarding only Customer.io would be inconsistent with the rest and isn't a reachable path for the supported SDK entrypoints. Out of scope here.

saikumarrs (approved):
I haven't checked the business logic changes but only focused on integration hygiene.

packages/analytics-js-integrations/src/integrations/CustomerIO/browser.js
```diff
@@ -28,45 +38,90 @@ class CustomerIO {
...
+    // until the client loads, so it is applied before any event.
+    const anonymousId = analytics.getAnonymousId?.();
+    if (anonymousId) {
+      window[V2_GLOBAL_NAME].setAnonymousId(anonymousId);
+    }
   }
 
   isLoaded() {
+    if (this.sdkVersion === SDK_V2) {
+      return window[V2_GLOBAL_NAME]?.initialized === true;
```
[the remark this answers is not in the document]
ItsSudip: `initialized` and `initialize` are two different members of the loaded client. Readiness here keys off `initialized` — a boolean declared `initialized = false` in the client (`core/analytics/index.ts:130`) and set to `true` after setup completes (`browser/index.ts:343`) — so it does flip to `true` on a real load; the "reports loaded on 2.x only once the client sets initialized" unit test covers that transition. `initialize` is a separate `@deprecated` no-op method (`init = this.initialize.bind(this)`), so `typeof …initialize === 'function'` would tie readiness to a deprecated no-op rather than the actual loaded state. The loader's `analytics.initialize || analytics.invoked` check is only the stub re-install guard, not the readiness signal. Keeping `initialized === true`.

packages/analytics-js-integrations/src/integrations/CustomerIO/browser.js
```diff
@@ -28,45 +38,90 @@ class CustomerIO {
...
+      // error, rather than silently loading nothing.
+      throw new Error(
+        'writeKey is required to load the Customer.io JavaScript client (SDK version 2.x); aborting load',
+      );
+    }
+    // Known-user in-app needs no client option (workspace setting). Anonymous in-app is an
+    // explicit opt-in, and it is the only reason to hand the In-App Plugin any options: the
+    // plugin's `siteId` comes from Customer.io's own settings for the write key, never from us.
+    const loadOptions = anonymousInApp
+      ? { integrations: { [IN_APP_PLUGIN_KEY]: { anonymousInApp: true } } }
```
maheshkutty, line 62: Are we not considering initializing colorSchema as part of these changes in this PR?

ItsSudip: No, we are not. We can add this as needed.

maheshkutty (approved):

maheshkutty:
@ItsSudip Lets record a video as well, which demonstrates in-app notification is working as expected

maheshkutty (approved):
LGTM

PR #3212, opened 2026-09-15
fix: stop DMT queue reclaiming XhrQueue items on storage quota error

### LINKED ISSUE #3211: BUG : DMT queue reclaims XhrQueue items after a localStorage quota error
**Describe the bug**

The DeviceModeTransformation queue and the XhrQueue use the same RetryQueue name, `rudder_<writeKey>`. When a localStorage write in the XhrQueue store throws QuotaExceededError, `Store.swapQueueStoreToInMemoryEngine` moves the store to memory. The swap has two defects. It writes the memory copy under the enum key (`IN_PROGRESS`) instead of the value (`inProgress`). It calls `this.remove(key)` with the enum key, so the localStorage copy stays behind.

The DMT queue then finds the XhrQueue `.ack` key in the shared memory engine, reads the stale localStorage copy through a LOCAL_STORAGE store, and reclaims XhrQueue items. Those items have no `destinationIds`. `sendTransformedEventToDestinations` throws:

```
TypeError: Cannot read properties of undefined (reading 'includes')
    at Array.filter
    at sendTransformedEventToDestinations
```

The throw happens inside the `getAsyncData` `.then`, the `.catch` calls the callback again, and the second throw is an unhandled rejection.

**To Reproduce**

Verified with Playwright and the 3.33.0 bundled ESM build.

1. Load the SDK with `queueOptions.batch.enabled: true` and one custom device-mode destination whose `shouldApplyDeviceModeTransformation` is true.
2. Send one track event, then make the next `localStorage.setItem` for the XhrQueue `batchQueue` key throw `QuotaExceededError`, and send a second event.
3. The SDK logs "Switching to in-memory storage". The old `rudder_<writeKey>.<id>.batchQueue` entry stays in localStorage with the first event.
4. Wait about 15 seconds. The DMT queue posts a third `/transform` request whose batch entry has no `destinationIds`, and the TypeError surfaces as an unhandled rejection.

The same harness against a bundle with the DMT queue renamed to `rudder_dmt` produces two `/transform` requests and no error.

**Expected behavior**

Each RetryQueue has a distinct name, so no queue reclaims another queue's items. The swap moves and removes the correct keys. `sendTransformedEventToDestinations` tolerates a missing `destinationIds`.

**Additional Information**

- SDK installation type: NPM, `@rudderstack/analytics-js/bundled`
- SDK version: 3.33.0 and 3.33.1; the code is unchanged on `develop`
- Framework: TanStack Start (React), bundled with Vite
- Relevant code: `deviceModeTransformation/constants.ts` (`QUEUE_NAME = 'rudder'`), `xhrQueue/constants.ts` (`QUEUE_NAME = 'rudder'`), `StoreManager/Store.ts` (`swapQueueStoreToInMemoryEngine`)

Co-authored-by: Claude Fable 5.1 <noreply@anthropic.com>

saikumarrs:
## PR Description

Fixes #3211.

The DeviceModeTransformation queue and the XhrQueue both resolved to the store name `rudder_<writeKey>`. They sit on different storage engines, so the collision stays invisible — until a localStorage quota error moves the XhrQueue store into the in-memory engine and merges the two keyspaces. DMT then reclaims XhrQueue items, which carry no `destinationIds`, and the resulting `TypeError` escapes as an unhandled rejection with the item wedged in `inProgress` forever.

DMT's queue is renamed `rudder_dmt`. It is memory-backed, so nothing persists across reloads; renaming XhrQueue instead would orphan events already sitting in localStorage from an installed 3.33.x.

The quota swap now derives keys through `createValidKey` and writes back through `set()`. It previously stored the memory copy under the enum key and never removed the durable copy, so migrated events were unreadable and the surviving stale copy is what DMT fed on. Client data stores keep their durable copy — dropping `rl_user_id` there would lose identity on the next page load.

`reclaim` reads the handle it handshook with instead of building a fresh localStorage store; without that, the now-working removal would delete another queue's events at quota.

Worth reading first: `Store.ts`, then `RetryQueue.reclaim`.

## Linear task (optional)

https://linear.app/rudderstack/issue/SDK-5473/rename-dmt-queue-fix-storage-swap-key-handling-and-the-dmt-callback

## Cross Browser Tests

- [x] Chrome
- [x] Firefox
- [x] IE11

## Sanity Suite

## Security

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented malformed destination data from being processed and added clearer error reporting.
  - Improved retry-queue recovery when storage switches between persistent and in-memory modes, helping prevent event loss.
  - Added safeguards against conflicting queues with the same name.
  - Corrected storage migration so queue data moves appropriately while client data remains preserved.
  - Improved asynchronous request error handling to prevent duplicate callbacks and unhandled errors.
  - Updated Device Mode Transformation queue naming to avoid conflicts.


packages/analytics-js-plugins/src/deviceModeTransformation/utilities.ts
```diff
@@ -79,6 +80,19 @@ const sendTransformedEventToDestinations = (
   const NATIVE_DEST_EXT_POINT = 'destinationsEventsQueue.enqueueEventToDestination';
   const ACTION_TO_SEND_UNTRANSFORMED_EVENT = 'Sending untransformed event';
   const ACTION_TO_DROP_EVENT = 'Dropping the event';
+
+  // Items that don't carry destination IDs don't belong to this queue
+  // and cannot be delivered to any destination
+  if (!Array.isArray(destinationIds)) {
```
saikumarrs: Do we still need this validation given that we're fixing the main queue issue?

packages/analytics-js/src/services/StoreManager/Store.ts
```diff
@@ -71,23 +71,31 @@ class Store implements IStore {
...
-      const validKey = noCompoundKey ? key : [name, id, key].join('.');
-
-      inMemoryStorage.setItem(validKey, value);
-      // TODO: are we sure we want to drop clientData
-      //  if cookies are not available and localstorage is full?
-      this.remove(key);
-    });
+    const existingEntries = Object.values(validKeys).map(
+      storeKey => [storeKey, this.get(storeKey)] as [string, any],
+    );
```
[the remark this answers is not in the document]
saikumarrs: Confirmed that `initClientDataStores` creates these stores without `validKeys`, so `validKeys` is `{}` and nothing migrates for them. That predates this PR though: the previous code iterated `Object.keys(validKeys)` over the same empty object, so the behaviour is unchanged either way.

What was misleading was the test, so I dropped `should migrate client data to the in-memory engine` - it passed `COOKIE_KEYS` explicitly, a configuration the product never creates. The retention test stays and now carries a note about the production shape, since it pins the `noCompoundKey` branch regardless of how the store is configured.

Client data being unreadable after a quota swap until the next page load is a real gap, just an orthogonal one to the queue collision this PR fixes. Tracking it separately rather than widening the diff into `StoreManager` behaviour.

packages/analytics-js-plugins/__tests__/utilities/retryQueue/RetryQueue.test.ts
```diff
@@ -623,6 +623,74 @@ describe('Queue', () => {
...
+    // tryReclaim's first write to the donor's handle is where the quota error lands
+    // in production. Emulate Store.set's quota path there: swap the handle to the
+    // in-memory engine and drop its durable copy.
+    const originalSetStore = defaultStoreManager.setStore;
+    let swapped = false;
+    defaultStoreManager.setStore = (config: any) => {
+      const created = originalSetStore(config) as Store;
+
+      if (config.id === donorId && !swapped) {
+        const originalSet = created.set;
```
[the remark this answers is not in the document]
saikumarrs: The mock `Store` this suite imports defines `set` as an arrow-function class property, not a prototype method, so it is lexically bound to the instance at construction. `const originalSet = created.set` keeps its receiver and `originalSet(key, value)` resolves `this.createValidKey` correctly.

The test does exercise the intended scenario: it fails against the unfixed `reclaim` with `processQueueCb` called 0 times, and passes once `reclaim` takes the store handle. Leaving it as-is.

packages/analytics-js/src/services/StoreManager/Store.ts
```diff
@@ -71,23 +71,31 @@ class Store implements IStore {
...
+
+    // Client data stores keep their durable copy. Dropping rl_user_id and the
+    // other cookie keys here would lose the identity on the next page load.
+    if (!noCompoundKey) {
+      existingEntries.forEach(([storeKey]) => this.remove(storeKey));
```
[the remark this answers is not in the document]
saikumarrs: Good point, reordered. The swap now writes each value to memory, confirms it reads back, and only then drops the key from the engine captured before the switch.

Added one guard beyond the suggestion: the removal is skipped when the captured engine is already the in-memory one, so a second swap on an already-swapped store cannot delete what it just wrote. The old remove-before-write order did not have that hazard.

packages/analytics-js-plugins/src/utilities/retryQueue/RetryQueue.ts
```diff
@@ -225,6 +238,12 @@ class RetryQueue implements IQueue<QueueItemData> {
       this.stop();
     }
 
+    const runningQueueId = runningQueueIdsByName.get(this.name);
+    if (runningQueueId && runningQueueId !== this.id) {
+      this.logger?.error(RETRY_QUEUE_NAME_COLLISION_ERROR(RETRY_QUEUE, this.name));
```
saikumarrs: We should block the queue at the initialisation level itself and not proceed forward.

packages/analytics-js-plugins/src/deviceModeTransformation/constants.ts
```diff
@@ -6,7 +6,7 @@ const DEFAULT_TRANSFORMATION_QUEUE_OPTIONS = {
 
 const REQUEST_TIMEOUT_MS = 10 * 1000; // 10 seconds
 
-const QUEUE_NAME = 'rudder';
+const QUEUE_NAME = 'rudder_dmt';
```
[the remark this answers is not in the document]
saikumarrs: The ambiguity is real in principle, but restructuring the name is the more dangerous change. The store name is the prefix of every persisted key (`<name>.<id>.<status>`), so encoding it as a tuple changes the keys for all four queues, including `rudder_beacon_*` and `rudder_destinations_events_*`. Events already sitting in localStorage from an installed 3.33.x would be orphaned on upgrade. That is exactly why this PR renames the DMT queue, which is memory-backed and persists nothing, rather than XhrQueue.

The collision it describes also needs a write key that literally begins `dmt_`, and write keys are generated opaque ids rather than user-chosen strings.

For the case that is genuinely reachable - two queues resolving to the same computed name from different call sites - `43383fd` adds a runtime guard: `RetryQueue.start` logs an error when another running queue already holds the name, and a unit test asserts the four plugin `QUEUE_NAME` constants stay distinct so a duplicate literal fails CI.

packages/analytics-js/src/services/StoreManager/Store.ts
```diff
@@ -71,23 +71,40 @@ class Store implements IStore {
...
+      // Write through set() so the values are serialised and keyed exactly as
+      // every later read expects them.
+      this.set(storeKey, value);
+
+      // Only give up the durable copy once the value is readable from memory:
+      // set() swallows serialisation and storage failures, and get() returns
+      // null on a parse or decryption failure.
+      const validKey = this.createValidKey(storeKey);
+      if (canDropDurableCopy && validKey && !isNullOrUndefined(this.get(storeKey))) {
+        durableEngine.removeItem(validKey);
```
dciccale: One edge case: if a reclaim-marker write hits the localStorage quota, the old queue moves into memory before recovery finishes. If the tab then pauses long enough, `ABANDON` cancels recovery. Later scans only search localStorage, so those events stay unsent and disappear on reload.

We could disable the memory fallback for these temporary recovery stores and skip the attempt when a marker write fails. The current tab’s own queue would still fall back to memory as it does today.

That would keep the saved events discoverable if recovery gets interrupted, without adding special handling for delayed timers. Delivery could wait until localStorage has room again, but a later attempt could still recover the events.

dciccale: Another edge case: this removal can throw `NS_ERROR_STORAGE_BUSY`, which the queue cleanup code already handles. In a fault test, that exception escapes `Store.set` and prevents the triggering write from reaching memory.

We could catch the removal error and continue migration, leaving that persistent entry for later cleanup. A cleanup failure would then stop interrupting the fallback that saves the event.

saikumarrs: Good catch, and you are right that passing the store handle into `reclaim` only covers the case where the handshake finishes. If `ABANDON` cancels it, the donor's entries are in memory on a handle nobody holds a reference to any more, and `findOtherQueues` only scans the reclaiming queue's own `getOriginalEngine()`, so nothing rediscovers them.

Went with the opt-out you suggested: `IStoreConfig` gains `noSwapOnQuota`, and the stores `findOtherQueues` builds for another queue set it. On a quota error those stores now stay on localStorage and `Store.set` returns without swapping, so the entries stay where a later scan can find them. The current tab's own queue keeps the fallback it has today.

No separate check on the marker write turned out to be needed: `createReclaimEndTask` already bails when `RECLAIM_START` does not read back as this queue's id, so a marker that fails to persist stops the handshake on its own.

saikumarrs: Confirmed and fixed. The swap runs from inside `Store.set`'s catch, so an `NS_ERROR_STORAGE_BUSY` from the removal escaped `set` entirely and the retry write never ran - losing the exact event the fallback exists to save. The removal is now wrapped and the stale entry is left for a later cleanup, which `RetryQueue.clearQueueEntries` already retries for this error.

Added a test that arms a quota error on the triggering write and a durable engine whose `removeItem` always throws, asserting the value still reads back from memory afterwards.

dciccale (approved):

